PR #105, opened 2016-03-10
Minor Changes in functions

atyamsriharsha:


ctjoreilly:
These changes won't be accepted. This is primarily because the code changes are more verbose and less clear. For example:

replacing:

return 0 == size();

with:

int size1 = size() ;
    if(size1==0)
        return true ;
return false ;

Is less clear, adds to the amount of code to be maintained and introduces unbraced conditionals, which are known to cause issues with code maintenance. Similarly, changing:

return poll();

to:

return this.poll();

is unnecessary as well as there is no ambiguity that the poll method belongs to the class/instance. Finally, changing:

if (offer(element)) {

to:

if(offer(element)!=false) {

is pointless and again makes the code unnecessarily verbose.

PR #110, opened 2016-03-11
TreeSearch closer to pseudocode

ghost:
Implements TreeSearch to follow pseudocode from Fig 3.7.


ctjoreilly:
Does not work in the context of the existing framework (i.e. does not compile). This is not an issue as new versions of the Search Algorithms have already been implemented on the AIMA4e branch that closely follow the pseudocode.

PR #113, opened 2016-03-11
Fixed Character encoding in the comments

atyamsriharsha:


ctjoreilly:
Won't accept, you are breaking legal HTML encodings, when the javadoc is generated these look correct as is.

Issue #35, opened 2015-03-12
Add implementation of - Fig 11.5  Hierarchical-Search


GoogleCodeExporter:
```
Fig 11.5  Hierarchical-Search.


```

Original issue reported on code.google.com by `ctjoreilly@gmail.com` on 23 Oct 2009 at 2:47

mshubhankar:
I have written a code for hierarchical search and also given its output in another file. The inputs are for now given as CLI string inputs for testing the program.

ctjoreilly:
Hi, when you have written unit tests for your code and are happy with how everything works please submit a pull request and I will review in detail.

Thanks

mshubhankar:
Actually, i have just tried to implement the algroithm and made the input menu driven by the user from the console.

mshubhankar:
This program takes input from problem.txt and input.txt which can be used as test case files.

ctjoreilly:
Ok,

But I still require a JUnit set of tests so testing can be automated. Can you also submit your proposed additions/changes via a push request (I will not be pulling in files individually from people). Also, look at how other code in the project is structured and commented. All submissions should follow a similar style.

Thanks

smsubham:
Tried to implement - Fig 11.5 Hierarchical-Search issue #35 . But on submission it says all test failed and no environment variables set. Please provide feedback so so that i can improve my code and better serve the community.

mshubhankar:
I am not able to find the REFINEMENT function . Is it written in the search framework ? or do i have to explicitely define it.

mshubhankar:
The algorithm contains a line APPEND(prefix,sequence,suffix), where all 3 of them are Actions. Could you please help me write this ?

smsubham:
Would it be ok if i try to implement last 2 lines of Pseudo Code i.e.
else for each sequence in REFINEMENTS(hla, outcome, hierarchy) do
frontier ← INSERT(APPEND(prefix , sequence, suffix), frontier)
in some other format.

samagra14:
This was resolved in #411. Hence, closing the issue.

PR #121, opened 2016-03-13
#35 Added the Code and Junit Test in the right place

mshubhankar:
Sorry for the delay in uploading the Junit Test file.
I have put the code in the nondeterministic folder.
As there is no class for a hierarchical problem , the code takes the problem through "problem.txt" and the HLAs through the "input.txt" file respectively. Please check if the algorithm is correct . 
Thanks


ctjoreilly:
The code as is does not compile. Also, the .txt files should be placed under and equivalent folder under resources - if just for testing then under the test folder's resources. Please fix and resubmit.

Thanks

Ciaran

Issue #122, opened 2016-03-13
AIMA4e implement Fig 4.2 Hill Climbing from AIMA3e branch


appoll:
Fig 4.2 Hill Climbing, page 122 in third edition. 
Issue needs AIMA4e label.

appoll:
I guess issue is not needed, since the pull request created one itself.

ctjoreilly:
I've been going back and forward between writing up explicit issues for each of the AIMA3e algorithms to be ported over to the AIMA4e branch. I think its probably a good idea just to make things more explicit. Perhaps a naming convention like: 'AMA4e rewrite of Fig 4.2 Hill Climbing from AIMA3e branch'. Feel free to enter. 

Thanks

Ciaran

appoll:
to be closed by pull request https://github.com/aimacode/aima-java/pull/123

PR #123, opened 2016-03-13
Rewrite implementation of Hill Climbing - closes issue #122

appoll:
After extended consultation with the existing code, I have finalized the Hill Climbing implementation for the AIMA4e branch. Tests are currently missing, so this feedback request comes as a need of confirmation of the existing implementation.

### DIFF
--- a/core/src/main/java/aima/core/api/search/local/HillClimbingSearch.java
+++ b/core/src/main/java/aima/core/api/search/local/HillClimbingSearch.java
@@ -67,6 +67,7 @@ public interface HillClimbingSearch<S> extends SearchFunction<S> {
         double h;
 
         SuccessorNode(Node<S> node, Function<Node<S>, Double> h) {
+            this.n = node;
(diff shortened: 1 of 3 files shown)


ctjoreilly:
Hi,

I've only had an opportunity to scan your code quickly but what is there looks good/promising. If you could work on a supporting JUnit set of tests for it as well that would be great. It will likely be a week or two before I can review in detail but keep up the good work.

Thanks

Ciaran

appoll:
Hello, 

I was fortunate enough to witness the benefits of writing tests and took advantage of the opportunity to fix some issues in the initial implementation. Tests are heavily based on the existing ones, but I still think there is room for improvement. Therefore I have also motivated some of my decisions / doubts in the comments, but these should be easily removable.

Greetings,

Paul

ctjoreilly:
Patch from pull request applied and modified to work with latest API changes.

PR #124, opened 2016-03-14
#35 Added the Code and Junit Test in the right place

mshubhankar:
Sorry there was a silly mistake the last time. I have tried to do it right this time.


ctjoreilly:
Hi,

Ok that's a start. Now look at the pseudo-code from the book and see how well your implementation maps to it. Ideally, follow the style of existing algorithms that have already been implemented that show the pseudo-code in the class comments and then structure the code so that its maps as best as possible to the pseduo-code. The algorithms itself for this issue shouldn't only work for a single example or have a main() associated with it. The related test cases should demonstrate how the general purpose algorithms should be used and its good to have the tests follow examples for the book. Also try to follow case and code layout similar to what is already implemented.

mshubhankar:
Ok, I am working on it sir. I am trying to make the code similar to the pseudo code in the book. Is the input style okay ? Is using a problem class like i have done here.. okay ?

ctjoreilly:
Closing.

PR #127, opened 2016-03-17
Implementation of Simulated-Annealing

anurag-rai:
Added implementation of the search SIMULATED-ANNEALING with J-Unit Tests.


ctjoreilly:
Looks very good. Will need to play with before bringing into the AIMA4e branch. Minor request, could you add a little documentation to the Scheduler class you implemented.

Thanks

anurag-rai:
Sure. I've added some comments as part of the documentation.

PR #128, opened 2016-03-17
Patch for Hierarchical Search

smsubham:
Attempt to implement Hierarchical Search along with junit unit tests to automate tests and proper documentation.


smsubham:
Hi
I have compiled it successfully using jdk 8 ,eclipse and all Junit 4.11 unit tests were also successful also passed ,but here it says all checks failed. Is there something else that i need to do to pass compilation test of Travis CI . Please point it out if any.
Thanks
Subham

anurag-rai:
Hi @smsubham, as you can see from the Travis-CI log
1> You have created two Problem.java classes:
- aima-core/src/main/java/aima/core/search/nondeterministic/HierarchicalSearch/Problem.java
- aima-core/src/main/java/aima/core/search/nondeterministic/problem.java

2> The issue is with the second path, where the filename is "problem.java", but is declared as "public class Problem" (Case Sensitive)

3> I'd recommend you use the already existing class for problem defined in:
package aima.core.search.framework;

smsubham:
Thanks for pointing it out. Just made an silly mistake while commiting files. Will fix it soon @anurag-rai

PR #142, opened 2016-03-20
Implementation of Simulated Annealing

anurag-rai:


ctjoreilly:
Applied patch manually making required updates with respect to replacement of Action interface with appropriate type parameter where necessary.

PR #145, opened 2016-03-21
Basic implementation of GeneticAlgorithm from Fig. 4.8

toohuman:
Just thought I'd get some feedback on the progress so far of implementing the GeneticAlgorithm interfaces and BasicGeneticAlgorithm class. Most of the code is based on that of the AIMA3e branch,  but hopefully the rewrite has stuck closer to the base algorithm from Figure 4.8 and Java 8 features have been added to take advantage of the more concise code, improving readability as a result.


ctjoreilly:
I would suggest moving the majority of the documentation (e.g. the class JavaDoc describing the algorithm) over to the interface. In addition, provide default implementations of the pseudo-code described in the interface should give you almost a 1 to 1 mapping between the two. The base implementation would then fill out a basic concrete implementation of the interface. Adding a set of JUnit tests is also a good thing to do.

Issue #130, opened 2016-03-20
AIMA4e reimplement Fig 4.11 And-Or-Graph-Search from AIMA3e branch


ctjoreilly:

anurag-rai:
The algorithm has been implemented, but the Test Suite needs working on. If anybody is keen to work on it please have a look at https://github.com/anurag-rai/aima-java/tree/AndOrSearch.

PR #146, opened 2016-03-21
Implementation of AndOrSearch #130

anurag-rai:
For #130 ,the interfaces for And and Or Search have been defined along with the base class. However, it lacks the Unit Test Suite. A look back at the AIMA3e branch showed that Classes from the Non-deterministic vacuum world were used to model the Test Cases. Should those Classes be re-implemented here too, or should there be a more generic model?


ctjoreilly:
Its nice to have tests that correspond to examples in the book so bringing over the non-deterministic vacuum world wouldn't hurt but not a requirement for the tests. Generic tests are fine too and as a baseline are likely better as they should cover more edge cases than an example based one.

core/src/main/java/aima/core/search/nondeterministic/Plan.java
```diff
@@ -0,0 +1,79 @@
...
+		return this;
+	}
+
+	/**
+	 * Returns the string representation of this plan
+	 * 
+	 * @return a string representation of this plan.
+	 */
+	@Override
+	public String toString() {
```
ctjoreilly, line 62: @anurag-rai 
Note: using JDK 8's StringJoiner class will let you write this method much more concisely:

anurag-rai:
@ctjoreilly I hadn't explored that feature. Re-wrote the part for it. Thanks for pointing it out. It sure saves a lot of code.

ctjoreilly:
Have decided to implement in a different way due to the recent changes in the aima4e search APIs.

Issue #131, opened 2016-03-20
AIMA4e reimplement Chp 4 Onine search problem from AIMA3e branch


ctjoreilly:

paarandika:
I would like to work on this issue. Is there any differences between 3e and 4e algorithm wise? Or is it re implementing same thing?

smsubham:
Hi
 AIMA4e branch is in progress , complete rewrite of the algorithms, leveraging JDK 8's new language features, from the AIMA3e branch. Algorithm would be same i suppose. Some part of AIMA4e is already implemented by @ctjoreilly  sir , i would recommend you to compare those implementation across AIMA3e and AIMA4e branch. @paarandika 

Regards
Subham

BenK10:
Well, there's another issue here. it looks like the AIMA4e branch also has differences in the core search and agent APIs. Without knowing about the design decisions behind that and whether they are complete, I'm not sure I can correctly implement any of the desired algorithms. In other words, it looks like there is more to it than just updating the algorithm itself to Java 8.

There is an open issue for adding a wiki for AIMA4e. That wiki would be the place to write about this.

BenK10:
Actually, I'm not sure there needs to be a separate OnlineSearchProblem for 4e because the Problem interface for 4e already supports all the behavior that 3e's OnlineSearchProblem has. In other words, since Problem is now more general than it used to be (compare 3e and 4e versions), it looks to me like you can use BasicProblem's implementation of it for an online problem, too. Or am I missing something?

ctjoreilly:
Have refactored the Problem interface to now be an aggregate of 5 explicit functional interfaces. The two variants of the Problem interface, NondeterministicProblem and OnlineSearchProblem aggregate their appropriate sets of functional interfaces. The BasicProblem implementation implements support for all 3 variants.

PR #147, opened 2016-03-21
Implementation of Online Search problem #131

anurag-rai:
Addresses Issue #131 (Complete Solution)
Complete Interfaces and classes implemented with working Test Suite.


anurag-rai:
@ctjoreilly As a design decision, I did not create interfaces for `ActionsFunction` and `GoalTest`. The ActionsFucntion is well handled by declaring a `Function<S,Set<Action>>` inside the OnlineSearch class and the GoalTest is also handled by declaring a `Predicate<S>` inside OnlineSearch.
I did create my own Functional Interface for `StepCostFunction` as it requires three arguments.

ctjoreilly:
Hi @anurag-rai , apologies for delay in getting back to you. Your design decisions sound good to me and I like that you are leveraging what is already provided as part of the JDK instead of re-inventing it.

ctjoreilly:
Functionality implemented slightly differently in recent commits.

Issue #132, opened 2016-03-20
AIMA4e reimplement Fig 4.21 Online-DFS-Agent from AIMA3e branch


ctjoreilly:

appoll:
Hi, 
I had already started tackling this one in the past week but at the moment I am stuck at writing the tests. However, if anyone is willing to add on top of it, the branch is [here](https://github.com/appoll/aima-java/tree/AIMA4e-132). 
Cheers, 
Paul

smsubham:
Hi Paul,
Ok let me try to write test cases for it.
Is implementation correct or do i need to work  on it too.
@appoll

appoll:
Well I wouldn't call it correct until it's tested. From my perspective it is close to being complete and if some things are missing they should be spotted when running the tests.

smsubham:
ok. Let me write test cases for it and try to get things right.

PR #207, opened 2016-03-22
Implementation of Online-DFS-Agent #132

anurag-rai:
Addresses and completely solves Issue #132 
A data-structure (`Two-key Hash-Map`) from the 3e branch has been utilized.
The Test Suite is a little lengthy as I didn't re-implement the `Map` and `Environment`classes, and it works for all different cases like:
- Working of the Class
- If Already at Goal State
- If No Path to Goal State
- A Normal Search
- Testing the Cases of Fig. 4.19 from the Book (3rd Edition)

P.S. I have used my own implementation of `OnlineSearchProblem` #131 which I've resolved in #147 
This is a further branch.


anurag-rai:
I was getting a `java.lang.VerifyError` and it was difficult to pin-point the dependency that was causing it. It was running fine in my system, but was failing over here. Finally managed to figure it out though.

ctjoreilly:
@anurag-rai out of curiosity what was the cause of the java.lang.VerifyError?

Does this push request overlap with #131 ?

anurag-rai:
@ctjoreilly In the J-unit Test class, I was declaring my `Graph`, `Actions-Function` and `Goal-Test-Function` inside the `Test` method, which threw the `java.lang.VerifyError`. As soon as I moved them out of the method and declared them as global, the error was resolved. Although, I didn't really understand the concept behind this.
The tests are working though. I've checked a lot of edge cases, scenarios and have run an example from the book too.

And yes, this push does overlap with #131 as the classes were needed.

[I've however sent a separate pull request(#147) for #131 alone as well]

ctjoreilly:
Have cherry picked code for inclusion in aima4e.

PR #208, opened 2016-03-23
CSP

mshubhankar:
Issue #136 
Added an interface to the CSP class and put CSP and all its dependencies to the core.search.csp package


ctjoreilly:
Implementing differently to correspond more closely with newer Search APIs.

PR #215, opened 2016-04-10
Update readme

smsubham:
Not sure if this is required. But this would help in avoiding confusion ,
as there are two implementation of FOLBCAsk in AIMA3e branch. Have corrected page no and specified 2nd implementation is for 2nd Edition of this book.


ctjoreilly:
There are often more than 1 implementation of any particular algorithm, no need to call out explicitly in the README in this instance. Also, this pull request includes an un-related file deletion.

PR #226, opened 2016-04-20
Implementation of AlphaBetaSearch for AIMA4e (#135)

anurag-rai:
Addresses #135 
Test Cases need implementation.


ctjoreilly:
Implemented slightly differently to conform with latest search APIs. Also fixed defect in main routine, which was not updating value of alpha across siblings of the root node.

PR #235, opened 2016-07-24
Implementing HITS Algorithm, with local Wikipedia pages as data

thundergolfer:
Implemented HITS Algorithm keeping pretty close to the pseudocode. Page data is local as requested (which is why this is 87K lines). 

Note: On account of either the data-set or a bug, the values tend toward 0 instead of converging. I am investigating this but still thought it worthwhile to merge this work. At least now others can see that it is implemented and they can comment, improve, and fix. Let me know if you think otherwise. 

Cheers


ctjoreilly:
Hi,
I will incorporate feedback directly into this push request. It may take me a few days to review fully/properly. The intent is that you consider the feedback given and then we'll work on incorporating your work into the main branch (will focus on your HITs  push request first and the CYK one later).

Thanks

Ciaran

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Artificial Intelligence A Modern Approach (3rd Edition): page 871.<br>
+ * <br>
+ * 
+ * <pre>
+ * function HITS(query) returns pages (with hub and authority numbers)
+ *   pages <- EXPAND-PAGES(RELEVANT-PAGES(query))
+ *   for each p in pages do
```
ctjoreilly, line 19: &larr; (i.e. & l a r r ;) can be used instead of <-

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+ * 
+ * <pre>
+ * function HITS(query) returns pages (with hub and authority numbers)
+ *   pages <- EXPAND-PAGES(RELEVANT-PAGES(query))
+ *   for each p in pages do 
+ *   	p.AUTHORITY <- 1
+ *   	p.HUB <- 1
+ *   repeat until convergence do
+ *   	for each p in pages do
+ *   		p.AUTHORITY <- SUM<sub>i</sub> INLINK<sub>i</sub>(p).HUB
```
ctjoreilly: &Sigma;<sub>i</sub> (i.e. &S i g m a;) can be used instead of SUM

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+ * RELEVANT-PAGES fetches the pages that match the query, and EXPAND-PAGES add in every page
+ * that links to or is linked from one of the relevant pages. NORMALIZE divides each page's
+ * score by the sum of the squares of all pages' scores (separately for both the authority
+ * and hubs scores.<br>
+ * <br>
+ * 
+ * @author Jonathon Belotti (thundergolfer)
+ *
+ */
+public class HITS {
```
ctjoreilly, line 40: All above looks good.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+    public double[] prevAuthVals;
+    public double[] prevHubVals;
+	public double prevAveHubDelta = 0;
+	public double prevAveAuthDelta = 0; 
+	////////////////////////////
+    
+    // TODO: Improve the convergence detection functionality
+    
+    public HITS( Hashtable<String,Page> pTable ) {
+    	this.pagesTable = pTable;
```
ctjoreilly: This comment applies to your usage of Hashtable throughout the code. Ideally you should use the interface Map<...,...>for all arguments and return types and not restrict it to the Hashtable implementation.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+ * and hubs scores.<br>
+ * <br>
+ * 
+ * @author Jonathon Belotti (thundergolfer)
+ *
+ */
+public class HITS {
+
+	final int RANK_HISTORY_DEPTH = 3;
+	final double DELTA_TOLERANCE = 0.05; // somewhat arbitrary
```
ctjoreilly: Ideally the above two values should be configurable through the constructor or setters.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+	Page[] pagesIndex;
+	Hashtable<String,Page> pagesTable;
+	// DETECT CONVERGENCE VARS
+    public double[] prevAuthVals;
+    public double[] prevHubVals;
+	public double prevAveHubDelta = 0;
+	public double prevAveAuthDelta = 0; 
+	////////////////////////////
```
ctjoreilly, line 50: Ideally the scope of these variables should not be 'public', package protected should be as wide a scope as you should permit for these.

ctjoreilly:
When I try to apply a patch for this pull request I get an issue with the file:

aima-core/src/main/resources/aima/core/ranking/data/pages/Ren/303/251_Descartes.txt

Where it gives me an error 'Unexpected exception while apply the patch. See log for a detailed error description. Reason: " is an invalid character in resource name '251_Descartes.txt"'

I believe this is the actual file:

 aima-core/src/main/resources/aima/core/ranking/data/pages/René_Descartes.txt 

causing the problem as its at this point that the patch fails (I am trying to apply on a windows machine).

Note: This is preventing me from applying the patch correctly as I do not see any of the test files in my local workspace after I see this error (i.e. it is only partially applying the patch).

Maybe the easiest thing to do is to create a pull request just for the resources i.e. (txt files and to rename the above file so it contains no special characters). So that we can cleanly merge those into the trunk first. Then we can have a separate pull request for just the Java code so that its easier to review/discuss.

Thanks

Ciaran

thundergolfer:
@ctjoreilly thanks for the feedback. I will look over it soon.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+    
+    // TODO: Improve the convergence detection functionality
+    
+    public HITS( Hashtable<String,Page> pTable ) {
+    	this.pagesTable = pTable;
+    }
+	
+	public List<Page> rank( String query ) {
```
ctjoreilly: I would recommend renaming this method to 'hits' and placing a commend above it:

// function HITS(query) returns pages with hub and authority number

so that its explicit that this is the function described in the pseudocode.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+    // TODO: Improve the convergence detection functionality
+    
+    public HITS( Hashtable<String,Page> pTable ) {
+    	this.pagesTable = pTable;
+    }
+	
+	public List<Page> rank( String query ) { 
+		List<Page> pages;
+		pages = expandPages( relevantPages( query, pagesTable ), pagesTable );
+		// for each p in pages
```
ctjoreilly, line 68: I would change the signatures on 'expandPages' and 'relevantPages' to not include 'pagesTable' as an argument. Instead, as pagesTable is a member attribute of the class I would just refer to it directly in those methods. The 2 main advantages of this are:
1. It more closely maps to the pseudocode.
2. Someone can override say theses two methods and instead use a different source for where the actual pages come from.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+		for( int i=0; i < pages.size(); i++ ) { 
+			Page p = pages.get(i);
+			p.authority = 1;
+			p.hub = 1; 
+		}
```
ctjoreilly, line 72: // for each page in pages do
for (Page p : pages) {
   p.authority = 1;
   p.hub = 1;
}

More closely maps to the pseudocode.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+		while( !convergence( pages ) ) {
+			// for each p in pages do
+			for( int i=0; i < pages.size(); i++ ) {
+				Page p = pages.get(i);
+				p.authority = SumInlinkHubScore( p );
+				p.hub       = SumOutlinkAuthorityScore( p );
+			}
```
ctjoreilly, line 78: recommend using for (Page p... looping construct instead and adding in comments where appropriate that map to the pseudocode.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+			normalize(pages);
+			try {
+				TimeUnit.SECONDS.sleep(1);
+			} catch (InterruptedException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
+			}
```
ctjoreilly: What is the purpose of the above sleep? Should this be removed?

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
```diff
@@ -0,0 +1,358 @@
...
+	 * Fetches and returns all pages that match the query
+	 * @param query
+	 * @return
+	 * @throws UnsupportedEncodingException 
+	 */
+	public List<Page> relevantPages( String query, Hashtable<String,Page> pTable ) {
+		List<Page> relevantPages = new ArrayList<Page>();
+		Iterator<String> it = pTable.keySet().iterator();
+		while(it.hasNext()) {
+			Page p = pTable.get(it.next());
```
ctjoreilly: A more concise version of this is just a for each on the values, i.e.:

for (Page p : pTable.values()) {
....

aima-core/src/main/java/aima/core/nlp/ranking/LinkFinder.java
```diff
@@ -0,0 +1,25 @@
...
+	
+	/**
+	 * Take a Page object and return its outlinks as a list of strings.
+	 * The Page object must therefore possess the information to determine
+	 * what it links to.
+	 * @param page
+	 * @return
+	 */
+	public List<String> getOutlinks( Page page );
+
```
ctjoreilly, line 16: There is no need to define 'public' on any methods or attributes on an interface as they are public by default.

aima-core/src/main/java/aima/core/nlp/ranking/Page.java
```diff
@@ -0,0 +1,33 @@
+package aima.core.nlp.ranking;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Page {
+	
+	public double authority;
+	public double hub;
```
ctjoreilly, line 9: It may make sense to add getters and setters for these 2 values instead of making them public. But as is works for a basic implementation.

aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
thundergolfer, line 72: I thought that syntax wasn't available in Java 7 and that I therefore shouldn't use it. I may be wrong about that though. It does map much better to the pseudocode

Edit: It appears I was very, very wrong about when the syntax was introduced. Will amend the code.

thundergolfer:
@ctjoreilly 

Thanks for the great code review. I have implemented everything and will resubmit the pull request as two pull requests. One for data, and one for code. 

Closing this PR

PR #236, opened 2016-08-18
HITS data. Data from Wikipedia pages

thundergolfer:
All the .txt file data for the HITS Algorithm to work with. 

This has been split off from a previous pull request, #235.


ctjoreilly:
Applied as patch and committed as part of https://github.com/aimacode/aima-java/commit/4b8143281d53e44c909900b251545dda88ceb0ef

Thanks

PR #237, opened 2016-08-18
Implementing HITS

thundergolfer:
Implemented HITS Algorithm keeping pretty close to the pseudocode. 

This pull request has been split off from the original pull request, #235. That pull request also included data for the HITS algorithm, which is now included in pull request #236. The code in this pull request has been cleaned up and refactored after advice from @ctjoreilly. 

Note: On account of either the data-set or a bug, the values tend toward 0 instead of converging. I am investigating this but still thought it worthwhile to merge this work. At least now others can see that it is implemented and they can comment, improve, and fix. Let me know if you think otherwise.

### DIFF
--- a/aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
+++ b/aima-core/src/main/java/aima/core/nlp/ranking/HITS.java
@@ -1,0 +1,364 @@
+package aima.core.nlp.ranking;
+
+import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


ctjoreilly:
Applied as patch and committed as part of: https://github.com/aimacode/aima-java/commit/4b8143281d53e44c909900b251545dda88ceb0ef

Thanks

PR #238, opened 2016-08-21
Implementation of the Monte-Carlo-Localization algorithm.

janphkre:
We have implemented the MCL algorithm as defined by the AIMA 3rd
edition pseudo code (figure 25.9). The MCL class is
aima.core.robotics.MonteCarloLocalization. All data types and other
classes are prescribed by interfaces. A map for a 2D environment is
implemented in the package aima.core.robotics.impl.map. A simple virtual
environment for the MCL algorithm can be found in
aima.gui.swing.demo.robotics.simple. The example application for that
environment is aima.gui.swing.demo.robotics.MonteCarloLocalizationApp.
The SVGGroupParserTest and the example application both load a map from
the resources. To get the maps included during the ant built both
build.xml files had to be adapted.

Please note that the JavaFX package prevented the successful build of
the aima-gui. We did not understand why a Java 8 GUI is included in an
otherwise Java 7 library. This would force anyone trying to build
the aima-gui to have java 8 installed besides java 7. We excluded the
JavaFX package from the ant build to resolve this inconsistency.


ctjoreilly:
Hi, I've only had a chance to scan through your pull request quickly but wanted to let you know that what you've put together looks like a great contribution to the aima-java project. At the moment I have limited time available so it may be a few weeks before I can integrate your work fully into the default branch. A few questions/comments:
1. As regards JDK 8 usage in aima-gui, this was a recent change, and for the time being we only plan to keep aima3e's aima-core module JDK7 compliant but if its causing confusion we may just allow the aima-core to upgrade to JDK 8 as well - do you see any issues with doing this?
2. It was not fully clear to me, though I've only looked at very quickly, how the code in MonteCarloLocalization maps to the pseudocode in Figure 25.9 - could you clarify further?
3. This is a very minor point but in a lot of your interfaces you include the 'public' access modifier to the method declarations - this is not necessary in interfaces as they are public by default.

Thanks

Ciaran

janphkre:
Hello,
I have updated the code according to your advice.
I did not see any problems upgrading both projects to JDK 8 and as far as I tested, everything is still working.
I extended the javadoc of the MonteCarloLocalization to be more clear about the pseudocode. If the code can still not be mapped to the pseudocode we may have to change the implementation a bit.

ctjoreilly:
Manually merged (very little changes just resolved a couple of minor conflicts).

PR #241, opened 2016-10-10
Bidirectional breadth first search 4e

globalworming:
parallel execution by utilizing streams

### DIFF
--- a/core/src/main/java/aima/core/environment/map2d/InState.java
+++ b/core/src/main/java/aima/core/environment/map2d/InState.java
@@ -1,6 +1,6 @@
 package aima.core.environment.map2d;
 
-public class InState {
+public class InState implements Comparable<InState> {
 	private String location;
 
 	public InState(String location) {
@@ -27,5 +27,17 @@ public class InState {
 	@Override
 	public String toString() {
 		return "In("+getLocation()+")";
+	}
+
+
+	@Override
+	public int compareTo(InState o) {
+		if (getLocation() == null) {
+			return -1;
+		}
+		if (o.getLocation() == null) {
+			return 1;
+		}
+		return getLocation().compareTo(o.getLocation());
 	}
 }
--- a/core/src/main/java/aima/core/search/api/DefinedGoalStatesProblem.java
+++ b/core/src/main/java/aima/core/search/api/DefinedGoalStatesProblem.java
@@ -1,0 +1,11 @@
+package aima.core.search.api;
+
+import java.util.List;
+
+/**
+ * we need a defined goal states for bidirectional problems
+ * @author wormi
+ */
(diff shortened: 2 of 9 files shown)


globalworming:
the readme asked for contributions, so I figured I could just try something from the ToDo list. I hope that's ok.

ctjoreilly:
Hi, No problem tackling from the TODO list. As the pull request is not self contained (i.e. it impacts existing code and is not just the algorithm) I'll need to look over in detail. Hopefully, within the next couple of weeks.

One request, it looks like you added logic to ProblemFactory.java but may have also reformatted the entire file as well (I need to look in a different diff tool to confirm). Ideally, its best to separate global format changes from code changes as it makes it difficult to pick out the meaningful changes.

Thanks

globalworming:
| Ideally, its best to separate global format changes from code changes as it makes it difficult to pick out the meaningful changes.
Your right. I will revert that and maybe separate the new code 

For the other changes comment on the PR and I will adjust as needed.

core/src/main/java/aima/core/environment/map2d/InState.java
```diff
@@ -1,6 +1,6 @@
 package aima.core.environment.map2d;
 
-public class InState {
+public class InState implements Comparable<InState> {
```
globalworming, line 3: needed for storing in concurrent set

core/src/main/java/aima/core/search/api/DefinedGoalStatesProblem.java
```diff
@@ -0,0 +1,11 @@
+package aima.core.search.api;
+
+import java.util.List;
+
+/**
+ * we need a defined goal states for bidirectional problems
+ * @author wormi
+ */
+public interface DefinedGoalStatesProblem<A, S> extends Problem<A, S> {
```
globalworming, line 9: I made it that we define the goal states. Otherwise we would have to explore the whole state space to find all states which satisfy the isGoalState() and use them as starting points for our backwards search.

ctjoreilly: I think the approach taken on the aima3e branch is a little clearer, see:

https://github.com/aimacode/aima-java/blob/AIMA3e/aima-core/src/main/java/aima/core/search/framework/problem/BidirectionalProblem.java

this lets you specify both starting points cleanly and to configure the sub-searches differently if needed. Have you reviewed aima3e's corresponding implementation and unit tests?

globalworming: saw it and forgot about it :) I just copy the BidirectionalProblem and update my code accordingly

PR #243, opened 2016-10-20
create metrics test

piyushgoel997:
created the unit test for the metrics class

But in the getInt,double,long functions the null case can't be tested as there is no way to set the value to be null (since primitive data types can't be null).


aima-core/src/test/java/aima/test/core/unit/search/framework/MetricsTest.class
```diff
@@ -0,0 +1,47 @@
...
+    public void before() {
+        metrics = new Metrics();
+    }
+
+    @Test
+    public void testGetInt() throws Exception {
+        int x = 893597823;
+        metrics.set("abcd",x);
+        assertEquals(123, metrics.getInt("abcd"));
+    }
```
ctjoreilly: This is a good start. However, I would recommend restructuring your tests. Instead of having a 'throws Exception' and having a JUnit assert check fail (you really want all the assertions to pass), I would test for the positive and negative cases with the different assert APIs. For example. the above could be written as:

@Test
public void testGetInt() {
     int x = 893597823;
     metrics.set("abcd",x);
     assertNotEquals(123, metrics.getInt("abcd"));
     assertEquals(x, metrics.getInt("abcd"));
}

piyushgoel997:
Thanks for reviewing my code and recommending improvements.
I have removed 'throws exception' and used assertNotEquals.
But the 'throws exception' part was auto generated that's why I didn't remove it earlier, the test never actually threw an exception.

ctjoreilly (commented):
Could you please rename the file to MetricsTest.java and ensure you can run it in JUnit.

piyushgoel997:
Renamed the file to MetricsTest.java and the tests are running properly in IntelliJ IDEA

ctjoreilly:
Manually merged, see:

https://github.com/aimacode/aima-java/commit/0c41f38936e4c2afbad88e56234f913e705a4d56

Only addition is inclusion of MetricsTest in the corresponding SearchTestSuite so that it gets included in the the full set of unit tests that are run.

Thanks

Ciaran

PR #244, opened 2016-11-06
minor speedup

manuel-delverme:
assuming the standard formulation of the 8puzzle problem, there can't be any two tiles with the same value


RuedigerLunde:
I like it :-)

PR #246, opened 2017-02-28
Update ReadMe

likhithlanka:
Just a small minor problem which i encountered. java should be installed in before hand.


ctjoreilly:
Will add correction by hand (due to change in file name).

PR #242, opened 2016-10-17
Bidirectional breadth first search 4e

globalworming:
copied BidirectionalProblem from AIMA3e

PR #249, opened 2017-03-01
Add Bidirectional Search Implementation

Manthan-R-Sheth:
The PR is aimed at implementing Bidirectional Search for the AIMA4e build.


Manthan-R-Sheth:
I shall be writing comments in the code and unit tests once the code is good to go.

Manthan-R-Sheth:
@ctjoreilly  if you could review this implementation.

ctjoreilly:
This overlaps with pull request #242 , so I need to look at both together.

ctjoreilly:
Also, could you include a Unit test and fully document your implemtation.

Thanks

Ciaran

Manthan-R-Sheth:
@ctjoreilly  ya i will do so and update the PR.
Thanks.

Manthan-R-Sheth:
@ctjoreilly i have updated the PR, if you can review and give your feedback. 
I have added unit test, and have extended BasicProblem with BidirectionalProblem to improve its functionality and allowing it to be generic as well.
Thanks.

core/src/main/java/aima/core/environment/support/ProblemFactory.java
```diff
@@ -13,10 +13,7 @@
 import java.util.Set;
 import java.util.stream.IntStream;
 
-import aima.core.environment.map2d.GoAction;
-import aima.core.environment.map2d.InState;
-import aima.core.environment.map2d.Map2DFunctionFactory;
-import aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania;
+import aima.core.environment.map2d.*;
```
globalworming: start imports: unnecessary change and bad practice IMHO
http://stackoverflow.com/a/147461/1359903
I'd revert that

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @Parameterized.Parameter
+    public String searchFunctionName;
+
+    public <A, S> List<A> searchForActions(Problem<A, S> problem) {
+        SearchForActionsFunction<A, S> searchForActionsFunction = new BidirectionalSearch<A, S>();
+        return searchForActionsFunction.apply(problem);
+    }
+
+    @Test
+    public void testBidirectionalSearch() {
```
globalworming: unittest tests 2 things, should be split into "start is goal" test and "from sibiu to fagaras" test

core/src/main/java/aima/core/environment/support/ProblemFactory.java
```diff
@@ -51,6 +48,21 @@
 				Map2DFunctionFactory.getStepCostFunction(simplifidRoadMapOfPartOfRomania));
 	}
 
+	public static Problem<GoAction, InState> getBidirectionalSearch(String initialState, final String goalLocations) {
```
globalworming: naming: this method does not return a BiDirectionalSearch as the name suggests but a BasicProblem

globalworming (changes requested):
unittests could test more cases e.g. zerind->sibiu and arad->neamt and arad->sibiu
use concrete imports instead of star imports

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
Manthan-R-Sheth: I though to do so, but according to the previous practise followed as seen [here](https://github.com/aimacode/aima-java/blob/AIMA4e/test/src/test/java/aima/test/unit/search/uninformed/BreadthFirstSearchTest.java), I combined the two tests into one.
@globalworming

Manthan-R-Sheth:
@globalworming 
Updated the PR.

core/src/main/java/aima/core/search/basic/support/BasicProblem.java
```diff
@@ -2,23 +2,17 @@
...
 
-import aima.core.search.api.ActionsFunction;
-import aima.core.search.api.GoalTestPredicate;
-import aima.core.search.api.NondeterministicProblem;
-import aima.core.search.api.OnlineSearchProblem;
-import aima.core.search.api.Problem;
-import aima.core.search.api.ResultFunction;
-import aima.core.search.api.ResultsFunction;
-import aima.core.search.api.StepCostFunction;
+import aima.core.search.api.*;
```
globalworming: Please Check your other changes for Star imports

Manthan-R-Sheth: Fixed the star imports. @globalworming, 
Actually Alt+Enter of IntelliJ does that. I will take care henceforth.

ctjoreilly:
I am having problems applying your patch, due to the multiple commits. Did you do any force commits to your pull request? Could I ask you to collapse all your changes into a single squashed pull request?

Manthan-R-Sheth:
@ctjoreilly
I will rebase and squash my commits and update the PR.

Manthan-R-Sheth:
@ctjoreilly
I have squashed my commits and updated the PR.

core/src/main/java/aima/core/environment/support/ProblemFactory.java
```diff
@@ -51,6 +51,21 @@
 				Map2DFunctionFactory.getStepCostFunction(simplifidRoadMapOfPartOfRomania));
 	}
 
+	public static Problem<GoAction, InState> getSimpleBidirectionalSearchProblem(String initialState, final String goalLocations) {
+		final SimplifiedRoadMapOfPartOfRomania simplifidRoadMapOfPartOfRomania = new SimplifiedRoadMapOfPartOfRomania();
```
ctjoreilly: goalLocations -> goalLocation, i.e. should be singular.

core/src/main/java/aima/core/search/basic/support/BasicProblem.java
```diff
@@ -102,4 +114,14 @@ public boolean isGoalState(S state) {
...
 	}
+
+	@Override
+	public Problem<A, S> getOriginalProblem() {
+		return this;
+	}
+
+	@Override
+	public Problem<A, S> getReverseProblem() {
+		return new BasicProblem<A, S>(finalState, initialState, goalTestPredicate, actionsFn, resultFn, stepCostFn);
```
ctjoreilly: More efficient to instantiate just once and keep as an attribute (see earlier comments).

core/src/main/java/aima/core/search/api/BidirectionalProblem.java
```diff
@@ -0,0 +1,11 @@
+package aima.core.search.api;
+
+/**
+ * Created by manthan on 2/3/17.
+ */
+
```
ctjoreilly: Change to '@author manthan'

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -0,0 +1,127 @@
...
+ */
+public class BidirectionalSearch<A, S> implements SearchForActionsFunction<A, S> {
+
+    @Override
+    public List<A> apply(Problem<A, S> problem) {
+        Node<A, S> node = newRootNode(problem.initialState(), 0);
+        if (problem.isGoalState(node.state())) {
+            return solution(node);
+        }
+        Node<A, S> revNode = newRootNode(((BidirectionalProblem<A, S>) problem).getReverseProblem().initialState(), 0);
```
ctjoreilly: TODO - we need to come up with a way to avoid this dirty cast, suggestions?

ctjoreilly: For now I would suggest adding a SearchForActionsBidirectionallyFunction interface similar to SearchForActionsFunction but that takes a BidirectionalProblem. Update all related code appropriately.

core/src/main/java/aima/core/environment/support/ProblemFactory.java
```diff
@@ -51,6 +51,21 @@
 				Map2DFunctionFactory.getStepCostFunction(simplifidRoadMapOfPartOfRomania));
 	}
 
+	public static Problem<GoAction, InState> getSimpleBidirectionalSearchProblem(String initialState, final String goalLocations) {
```
ctjoreilly: Should return a BidirectionalProblem.

core/src/main/java/aima/core/search/api/BidirectionalProblem.java
ctjoreilly: Remove blank line.

core/src/main/java/aima/core/search/basic/support/BasicProblem.java
```diff
@@ -1,24 +1,26 @@
...
 
 /**
  * Basic implementation of the Problem, NondeterministicProblem, and
  * OnlineSearchProblem interfaces.
  *
  * @author Ciaran O'Reilly
  */
-public class BasicProblem<A, S> implements Problem<A, S>, NondeterministicProblem<A, S>, OnlineSearchProblem<A, S> {
+public class BasicProblem<A, S> implements Problem<A, S>, NondeterministicProblem<A, S>, OnlineSearchProblem<A, S>, BidirectionalProblem<A, S> {
 	private S initialState;
```
ctjoreilly: Declare implements BidirecationalProblem after Problem<A, S>, i.e.:

implements Problem<A, S>, Bidirectional....

so that is matches the order these interfaces are introduced in the book.

core/src/main/java/aima/core/search/basic/support/BasicProblem.java
```diff
@@ -1,24 +1,26 @@
...
+	private S finalState;
```
ctjoreilly: change this to:

private Problem<A, S> reverseProblem and instantiate in the Bidirectional constructor, so that it only needs to be done so once.

core/src/main/java/aima/core/search/basic/support/BasicProblem.java
```diff
@@ -73,6 +75,16 @@ public BasicProblem(ActionsFunction<A, S> actionsFn, GoalTestPredicate<S> goalTe
 		this.stepCostFn = stepCostFn;
 	}
 
+	// BidirectionalProblem constructor(Reverse Problem)
```
ctjoreilly: Move this to below the problem constructors (and ensure you have a blank line between method definitions). Again so that the order matches that of the book in terms of introduction to concepts.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -0,0 +1,127 @@
...
+ * Artificial Intelligence A Modern Approach (4th Edition): Figure ??, page ??.
+ * <br>
+ * <br>
+ *
+ * <pre>
+ * The strategy of this search implementation is inspired by the description of
+ * the bidirectional search algorithm i.e. Problem is reversed and search is performed from both
+ * frontiers.
+ * Created by manthan on 1/3/17.
+ */
```
ctjoreilly: Change to '@author manthan'

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -0,0 +1,127 @@
...
+                List<A> sol = solution(solutionNode);
+                Collections.reverse(sol);
+                return sol;
+            }
+            solutionNode = pathExistsBidirectional(back, exploredBack, exploredFront, problem);
+            if (solutionNode != null) {
+                List<A> sol = solution(solutionNode);
+                Collections.reverse(sol);
+                return sol;
+            }
```
ctjoreilly, line 60: Is there a cleaner way to do this without repeating the conditional block twice?

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -0,0 +1,127 @@
...
+                    }
+                    exploredFront.add(child.state());
+                    queue.add(child);
+                }
+            }
+        }
+        return null;
+    }
+
+    //Supporting Code
```
ctjoreilly: Minor tidy up, change to:
//
// Supporting Code

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -0,0 +1,127 @@
...
+        return new HashSet<>();
+    }
+
+    public BidirectionalSearch() {
+
+    }
+
+    public boolean containsState(Queue<Node<A, S>> frontier, S state) {
+        return frontier.stream().anyMatch(frontierNode -> frontierNode.state().equals(state));
+    }
```
ctjoreilly: May want to consider a more efficient way of doing this. Have a look at BasicFrontierQueue, it will let you do a containment check much faster.

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
```diff
@@ -0,0 +1,49 @@
...
+        Assert.assertEquals(Arrays.asList((String) null),
+                searchForActions(ProblemFactory.getSimpleBidirectionalSearchProblem(
+                        SimplifiedRoadMapOfPartOfRomania.ARAD, SimplifiedRoadMapOfPartOfRomania.ARAD)));
+        //Result will contain all intermediate nodes excluding initial and final nodes.
+        Assert.assertEquals(
```
ctjoreilly: // NOTE: The answer should include the goal as well (i.e. Bucharest) otherwise the list of actions doesn't actually get you there. This should be fixed. I would also consider adding some additional tests.

ctjoreilly (changes requested):
Hi, 

This looks good as an initial version of a core implementation of the BidirectionalSearch algorithm. Could you please go through the review comments and make the appropriate changes and the re-push (again all squashed together as I apply as a patch).

Thanks

Ciaran

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
Manthan-R-Sheth: @ctjoreilly I will have to change ```NodeFactory<A,S>``` for making a new child node since now the problem is ```BidirectionalProblem``` and not Problem
In my opinion, ```Node<A, S> newChildNode(Problem<A, S> problem, Node<A, S> parent, A action)``` should besuch that it includes both types of problems.

Manthan-R-Sheth: Can I have ```Node<A, S> newChildNodeForBidirectionalSearch(BidirectionalProblem<A, S> problem, Node<A, S> parent, A action)``` as a function of ```NodeFactory<A,S>``` ?

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
Manthan-R-Sheth: ```front``` and ```back``` can be made ```BasicFrontierQueue<A,S>``` . It is good way for improving efficiency.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
Manthan-R-Sheth: I am not sure about an alternative way for this. This is because alternate evaluations of both queues need to be done for the algorithm.

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
Manthan-R-Sheth: Since the output is a list of Actions ( ```List<A>``` ), the goal has a null action and so I didn't add it to the result.

Manthan-R-Sheth:
@ctjoreilly if you could review now. 
**Changes made** :- 
1. SearchForActionsBidirectionallyFunction is implemented that takes BidirectionalProblem.
1. newChildNodeForBidirectionalSearch is added to make nodes for bidirectional search.
1. Queues are converted to BasicFrontierQueue for efficient state containment. (can be extended to other classes too).
1. reverseProblem attribute is added in BasicProblem.
1. Unit tests are increased.
1. Minor fixes of @author and new lines.

core/src/main/java/aima/core/search/api/BidirectionalProblem.java
```diff
@@ -0,0 +1,20 @@
...
+        InitialStateFunction<S>,
+        // List<A> actions(S s)
+        ActionsFunction<A, S>,
+        // List<S> results(S s, A a)
+        ResultFunction<A, S>,
+        // boolean isGoalState(S s)
+        GoalTestPredicate<S>,
+        // double stepCost(S s, A a, S s')
+        StepCostFunction<A, S> {
+    Problem<A,S> getOriginalProblem();
```
ctjoreilly: Why are you doing the above? Ideally, I'd like to keep this the way it was. I would like to also avoid changing the NodeFactory to take this into account could you explain further?

ctjoreilly (changes requested):
See review comment related to BidirectionalProblem interface. Also, this pull request has jumped to 13 files. Could you create a new pull request that only conains your changes (and not those also pulled from the aima4e branch).

Thanks

core/src/main/java/aima/core/search/api/BidirectionalProblem.java
Manthan-R-Sheth: @ctjoreilly 
I did this because :-

1. newChildNode() function requires a ```Problem```, but we have ```BidirectionalProblem``` as the argument of apply(). Hence to create a node for such case, I modified it accordingly. Could you suggest a better approach.
1. If BidirectionalProblem isn't implemented as above, we won't have functions like ```initialState()```. Thus to make it just like other types of problems like ```Problem```,
``` NondeterministicProblem``` etc I did so.

**Comment by Manthan-R-Sheth** [comment]
@ctjoreilly  I have rebased onto my commits and git history is clean.

File: core/src/main/java/aima/core/search/basic/support/BasicProblem.java [hunk 19]
```diff
@@ -1,24 +1,26 @@
...
 /**
  * Basic implementation of the Problem, NondeterministicProblem, and
  * OnlineSearchProblem interfaces.
  *
  * @author Ciaran O'Reilly
  */
-public class BasicProblem<A, S> implements Problem<A, S>, NondeterministicProblem<A, S>, OnlineSearchProblem<A, S> {
+public class BasicProblem<A, S> implements Problem<A, S>, BidirectionalProblem<A, S>, NondeterministicProblem<A, S>, OnlineSearchProblem<A, S> {
 	private S initialState;
+	private S finalState;
```
ctjoreilly: You shouldn't need to track the finalState if you've already instantiated the reverseProblem as I don't see where it will be used otherwise.

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
ctjoreilly: Thinking about this some more, I think we may need to change the interface on SearchForActionsBidirectionallyFunction to return something other than a List<A>. Don't consider addressing this until we get a basic version merged into the trunk that we agree upon and then we can come back round to see if this idea make sense:

Instead it returns something like this:

 interface BidirectionalActions<A> {
      List<A> fromInitialStatePart()
      List<A> fromGoalStatePart()

      List<A> fromInitialStateToGoalState(); // What I was indicating above I was expecting as  a result
      List<A> fromGoalStateToInitialState(); // essential the reverse of the previous method.
}

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
ctjoreilly: Thinking about this some more, I think we may need to change the interface on SearchForActionsBidirectionallyFunction to return something other than a List<A>. Don't consider addressing this until we get a basic version merged into the trunk that we agree upon and then we can come back round to see if this idea make sense:

Instead it returns something like this:

 interface BidirectionalActions<A> {

      List<A> fromInitialStatePart()
      List<A> fromGoalStatePart()
      List<A> fromInitialStateToGoalState(); // What I was indicating above I was expecting as  a result
      List<A> fromGoalStateToInitialState(); // essential the reverse of the previous method.
}

Manthan-R-Sheth: Ya i will surely update it when things get finalised.

core/src/main/java/aima/core/search/basic/support/BasicProblem.java
Manthan-R-Sheth: Actually it used to initialise the reverse problem , in the constructor.
```		this.reverseProblem = new BasicProblem<>(finalState, actionsFn, resultFn, goalTestPredicate, stepCostFn);```
It acts as the initial state for the reverse problem.
I will remove the global variabe though.

**Comment by Manthan-R-Sheth** [comment]
@ctjoreilly  Now is it good to go.

File: core/src/main/java/aima/core/search/api/BidirectionalProblem.java (hunk 18 above)
Comment (ctjoreilly) [reply core/src/main/java/aima/core/search/api/BidirectionalProblem.java]: I thought I'd submitted a response as regards this earlier. I do not wish to go with the BidirectionalProblem API in this form or require any changes to the NodeFactory. Could you explain to me the difference between the results from call BidrectionalProblem.initialState() and BidirectionalProblem.getOriginalProblem().initialState()? Similarly, for the other overlapping methods. Basically, I only see BidirectionalProblem needing to be a container for the original and reverse versions of the problem, it itself does not need to be one. Let me know if I'm missing something here.

Thanks

Ciaran

Comment (Manthan-R-Sheth) [reply core/src/main/java/aima/core/search/api/BidirectionalProblem.java]: BidirectionalProblem.initialState() and BidirectionalProblem.getOriginalProblem().initialState(), both will have same results. I made BidirectionalProblem.getOriginalProblem() for this purpose only. But I used it as a different problem to keep continuity with the superclasses of BasicProblem (which are all problems in themselves). 
Shall I replace BidirectionalProblem.getOriginalProblem().initialState() ? Your call on this aspect.

@ctjoreilly

**Comment by ctjoreilly** [comment]
Hi,

To avoid confusion with the semantic meaning of BidireactionalProblem I think the simplest thing to do is delete it and revert BasicProblem and NodeFactory back to the way they were. Instead modify the apply() method on SearchForActionsBidirectionallyFunction to:

List<A> apply(Problem<A, S> originalProblem, Problem<A, S> reverseProblem);

and the method on ProblemFactory should now return:

Pair<Problem<GoAction, InState>, Problem<GoAction, InState>> getSimpleBidirectionalSearchProblem(...)

This will also help with reducing API bloat. Once done and all tests passing I think this should be good for merging to AIMA4e as the basis for further improvements on BidirectionalSearch.

Thanks

Ciaran

File: core/src/main/java/aima/core/search/api/SearchForActionsBidirectionallyFunction.java [hunk 20]
```diff
@@ -1,13 +1,15 @@
...
 
 /**
  * @author manthan
  */
 @FunctionalInterface
-public interface SearchForActionsBidirectionallyFunction<A, S> extends Function<BidirectionalProblem<A, S>, List<A>> {
+public interface SearchForActionsBidirectionallyFunction<A, S> extends Function<Pair<Problem<A, S>, Problem<A, S>>, List<A>> {
     @Override
-    List<A> apply(BidirectionalProblem<A, S> problem);
+    List<A> apply(Pair<Problem<A, S>, Problem<A, S>> problem);
```
ctjoreilly: This should not take a pair, but broken apart as:

originalProblem, reverseProblem

Using Pair on the ProblemFactory was for convenience as you can't specificy generics in an Array i.e. Problem[].

Manthan-R-Sheth:
@ctjoreilly  Reverted back with BasicProblem and NodeFactory.
Updated accordingly.

ctjoreilly:
Could you squash/rebase so I can validate the patch before accepting the pull request (almost there :) ).

Thanks

Ciaran

Manthan-R-Sheth:
Actually I had rebased locally earlier but forgot to push it on my fork. And I made changes on it locally. I had to merge after that. I cleaned that history by cherrypicking but now I am not able to squash my commits. 
if you could test my PR locally by ```git fetch origin pull/ID/head:BRANCHNAME``` instead of applying patch.
Thanks.

ctjoreilly:
Ok, I'll chance a direct merge as I think it should be good (hopefully)

Manthan-R-Sheth:
@ctjoreilly 
Thanks.

PR #250, opened 2017-03-02
Optimise GraphPriorityQueueSearch by using Priority Queue 

Manthan-R-Sheth:


ctjoreilly:
This is failing to compile can you please look into.

Manthan-R-Sheth:
It is failing on the TreePriorityQueueSearch class' implementation.  There are two ways to overcome this :-
1. Converting tree search to priority queue implementation.
1. Using a flag to give queue and priority queue respectively for graph and tree algorithms.

First one would be a better implementation in my opinion but some advise in this direction would be really helpful.

ctjoreilly:
I am going to close this pull request. If you would like to resubmit, could you do the following:

1. Ensure you do not reformat the entire file when you modify it. It makes it difficult to view the differences.

2. Write up the goals for your pull request.

Thanks

Ciaran

Issue #193, opened 2016-03-22
AIMA4e implement Fig 18.8 Cross-Validation-Wrapper from AIMA3e branch


ctjoreilly:

shashvat-kedia:
I would like to work on this.

PR #253, opened 2017-03-08
issue #193 18.8 Cross Validation Wrapper implementation 

Allen-Hao-Huang:
CrossValidationWrapper() is not guaranteed to terminate as the psuedo code is from 1 to inf

Check for convergence is checking if 2 values are very close to each other, however it still might not mean it has converged

### DIFF
--- a/aima-core/src/main/java/aima/core/learning/framework/LearnHypothesis.java
+++ b/aima-core/src/main/java/aima/core/learning/framework/LearnHypothesis.java
@@ -5,15 +5,15 @@ import aima.core.learning.knowledge.Hypothesis;
 /**
  * made by allen
  */
-public interface Learn_Hypothesis {
+public interface LearnHypothesis {
 
     /**
      * @param complexity the complexity for the model of the hypothesis
-     * @param training_sample the dataset for the hypothesis to be based on
+     * @param trainingSample the dataset for the hypothesis to be based on
      *
      * @return returns hypothesis
      */
-    Hypothesis make_hypothesis(int complexity, DataSet training_sample);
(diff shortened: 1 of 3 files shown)


ctjoreilly:
Hi,

This is a good start. I would ask that you review the general guidelines here:

https://github.com/aimacode/aima-java/wiki/AIMAJava-Contributing

In particular, could you:

1. Include JUnit tests demonstrating the algorithm working (should run quickly, can just be simple example).
2. Include JavaDoc pseudocode of the algorithm in the class comment block, see:

https://github.com/aimacode/aima-java/blob/AIMA3e/aima-core/src/main/java/aima/core/environment/vacuum/ReflexVacuumAgent.java

for an example.

As regards the not guaranteed to terminate, that is fine, this is just intended to be a simple example and adding a NOTE at the bottom of the class javadoc highlighting this is sufficient (addressing it can be done by with an alternative implementation that addresses/improves upon the pseudocode but that should be a separate submission).

Thanks

Ciaran

Allen-Hao-Huang:
thanks for the feedback! I will try get it done soon

Allen-Hao-Huang:
@ctjoreilly currently writing JUnit tests with a polynomial regression model as the learner, how would you store paired data values into the class Example for the dataset? is targetValue() inside Example.java supposed to return a value from attributes?

ctjoreilly:
Have a look at DataSetFactory.getIrisDataSet(), this creates a dataset from file with numeric attributes. I think this should be close to what you need. Also, did you intentionally close this issue?

Allen-Hao-Huang:
@ctjoreilly issue 43/193 is still there, i don't think i closed the issue? This was a pull request and i closed it because i'm writing the JUnit tests

PR #254, opened 2017-03-09
implement a bidirectional search

globalworming:


globalworming:
rebased, squashed, as requested in #242

core/src/main/java/aima/core/environment/map2d/InState.java
```diff
@@ -28,4 +28,16 @@ public int hashCode() {
...
+
+
+	@Override
+	public int compareTo(InState o) {
+		if (getLocation() == null) {
+			return -1;
+		}
+		if (o.getLocation() == null) {
+			return 1;
+		}
```
ctjoreilly, line 40: Both of these are conditions that should cause a NullPointerException to be thrown, see:
https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html

ctjoreilly: An alternative way of doing this to making InState Comparable would be to create a Comparator like this:

Comparator.<InState>nullsFirst(Comparator.comparing(is -> is.getLocation()));

Which I think makes a little more sense in this case.

core/src/main/java/aima/core/search/api/BidirectionalProblem.java
```diff
@@ -0,0 +1,16 @@
...
+ * once (i.e InitialState<->Goal).
+ *
+ * @author Ciaran O'Reilly
+ *
+ */
+public interface BidirectionalProblem<A, S> {
+  Problem<A, S> getOriginalProblem();
+
+  List<Problem<A, S>> getReverseProblems();
+}
```
ctjoreilly, line 16: Bidirectional implies functioning in two directions. This could be > 2, which is not what we want. It should be Problem<A, S> getReverseProblem().

core/src/main/java/aima/core/search/basic/support/BasicNode.java
```diff
@@ -42,14 +42,17 @@ public A action() {
...
-		sj.add(parent == null ? "null" : parent.state().toString());
-		sj.add(action == null ? "null" : action.toString());
-		sj.add(""+pathCost);
+		StringJoiner sj = new StringJoiner(", ", "{", "}");
+		sj.add("parent: " + (parent == null ? "null" : parent.toString()));
+		sj.add("state: '" + state.toString() + "'");
+		sj.add("fromAction: " + (action == null ? "null" : "'" + action.toString() + "'"));
+		sj.add("withPathCost: "+pathCost);
 		return sj.toString();
 	}
```
ctjoreilly, line 58: I do not wish to add JSON formatting of toStrings() within the core package. Ideally, an output specific format should be handled in a separate way.

core/src/main/java/aima/core/environment/support/BidirectionalRomaniaProblem.java
```diff
@@ -0,0 +1,118 @@
...
+  @Override
+  public List<Problem<GoAction, InState>> getReverseProblems() {
+    List<Problem<GoAction, InState>> problems = new LinkedList<>();
+    for (String goalLocation : goalLocations) {
+      problems.add(new MyReverseProblem(goalLocation));
+    }
+    return problems;
+  }
+
+
```
ctjoreilly, line 55: I am not sure why you need these two custom classes below (MyOriginalProblem and MyReverseProblem). You could just use:
aima.core.search.basic.support.BasicProblem
and configure it appropriately for each usage.

core/src/main/java/aima/core/search/basic/support/BasicNode.java
```diff
@@ -59,7 +62,7 @@ public String toString() {
 		this(state, null, null, pathCost);
 	}
 
-	BasicNode(S state, Node<A, S> parent, A action, double pathCost) {
+	public BasicNode(S state, Node<A, S> parent, A action, double pathCost) {
```
ctjoreilly, line 65: This should not be 'public' as the intent is to only allow nodes to be instantiated via a NodeFactory (in this case a BasicNodeFactory).

core/src/main/java/aima/core/search/basic/support/BidirectionalSearch.java
```diff
@@ -0,0 +1,306 @@
+package aima.core.search.basic.support;
+
```
ctjoreilly, line 2: This is located in the wrong package, the description in the book is described in the context of uninformed search algorithms, and therefore the version included in the 'core' module should adhere to this restriction. A basic Bidrectional implementation should go in:

aima.core.search.basic.uninformed

However, more advanced versions could be added to the 'extra' module as alternative implementations. That either exhibit more functionality or some other plus over the basic description in the book. Note: I am under the impression that the Bidirectional search description will change in AIMA4e based on recent research on the approach (TBD).

core/src/main/java/aima/core/search/basic/support/StateActionTimeLine.java
```diff
@@ -0,0 +1,31 @@
+package aima.core.search.basic.support;
```
ctjoreilly, line 1: This type of utility is likely best not added to the 'core' module. The 'extra' module is likely better but I would recommend removing this kind of thing from an initial submission as it bloats the functionality prematurely.

core/src/main/java/aima/core/environment/support/ProblemFactory.java
```diff
@@ -51,7 +52,24 @@
 				Map2DFunctionFactory.getStepCostFunction(simplifidRoadMapOfPartOfRomania));
 	}
 
-	public static Problem<String, VEWorldState> getSimpleVacuumWorldProblem(String inInitialLocation,
+  public static BidirectionalProblem<GoAction, InState> romaniaRoadMapProblem(
```
ctjoreilly, line 55: I would name this method:

getSimplifiedRoadMapOfPartOfRomaniaBidirectionalProblem()

to keep it consistent with the method defined above it.

core/src/main/java/aima/core/search/basic/support/StraightLineDistanceHeuristic.java
```diff
@@ -0,0 +1,42 @@
...
+import aima.core.environment.map2d.Map2D;
+import aima.core.search.api.Node;
+import aima.core.util.datastructure.Point2D;
+
+import java.util.Arrays;
+import java.util.function.ToDoubleFunction;
+
+/**
+ * @author wormi
+ */
```
ctjoreilly, line 14: This is specific to 'GoAction' and 'InState' so does not belong within the basic.support package. To re-use I would recommend moving it from InformedSearchTest to ProblemFactory, and declaring it as a public static class (I will likely make this change).

test/src/test/java/aima/test/unit/environment/map2d/PrintRomaniaMap.java
```diff
@@ -0,0 +1,66 @@
+package aima.test.unit.environment.map2d;
```
ctjoreilly, line 1: This is kind of neat. Would it be possible to add this as a separate standalone pull request whereby:
1. It is located in aima.core.environment.map2d
2. Is generalized to printing an Map3D and renamed to PrintMap2D
3. Include a main(String[])  and updates inner methods to support specifying where the output should be written.

If this has any dependencies on third party libraries (I don't think it does) then it should reside in the 'extra' module and have its package named updated to reflect this.

test/src/test/java/aima/test/unit/search/basic/BasicNodeTest.java
```diff
@@ -0,0 +1,38 @@
+package aima.test.unit.search.basic;
```
ctjoreilly, line 1: No need for this to be part of a BidirectionalSearch pull request. This is a general test but won't be adding as it is based on JSON toString() output expectations.

ctjoreilly (commented):
Hi, 

Some general feedback. 

1. This is too big of a pull request (part of the reason its taken me so long to go through it in detail). In the future I recommend focusing on a single self contained piece of functionality (e.g. just an initial version of BidirectionalSearch with a basic JUnit test, avoid changing existing supporting code and classes if at all possible).

2. I will not be merging this pull request, as I think the functionality requires some more thinking. I would focus on an uninformed implementation first instead of jumping to the more complex informed case. I would also consider using a different mechanism for selecting different strategies. The queue used is only part of the strategy, in particular when you are trying to take informed search into account as well.

With all that said, thanks for your work on this, its a tricky problem! In addition, part of me wants to hold off on Bidirectional Search implementations for AIMA4e until a clearer picture of what will be included in the new edition of the book is known.

globalworming:
ok, instead of replying to everything, I will just create new pull requests. Some of the criticism is valid, others will maybe get resolved when we talk about it in smaller pull requests.

I will start with a PR for BiDirectionalBreadthFirstSearch with a single goal state (hopefully I can tone down my feature set before @Manthan-R-Sheth gets his PR approved). I'd really like to see my other features (multiple goalstates, A-Star..) included. I put some thought into it and was kinda proud to be able to say "forward search is A*, backward search is uniform cost search, give me whatever brings me to A or B". If that's more than you'd want to include.. well my bad I guess. We'll see..

regards
globalworming

ctjoreilly:
Hi,

I am good with the enhancement ideas you have (they are cool, I would also encourage you to look in detail at the version of BidirectionalSearch in AIMA3e as I think the capabilities overlap a bit). For > 2 search directions though I would consider adding a different algorithm (interface) called NaryDirectionalSearch, which can go from 2 or more states. This would be included as part of the 'extra' module as a generalization of the bidirectional idea. As regards, @Manthan-R-Sheth 's version it will likely be merged into AIMA4e as the basis for the 'core' version of the algorithm as it most closely matches what is discussed in the book (i.e. the implementation of the algorithms in 'core' are to match the book as best as possible while those in 'extra' can diverge as much as is sensible, which is more where your contribution sits).

Thanks

Ciaran

PR #257, opened 2017-03-10
Add Priority Queue to GraphPriorityQueueSearch and TreePriorityQueueSearch

Manthan-R-Sheth:
Goals :-
1. Priority queue improves readability for graph and tree algorithms as mentioned in the description of the algorithm.
1. Prevents many calls to function ``getNodeComparator().compare(child, n)`` in `` removedNodeFromFrontierWithSameStateAndLowerPriority() ``.

### DIFF
--- a/extra/src/main/java/aima/extra/search/pqueue/AbstractQueueSearchForActions.java
+++ b/extra/src/main/java/aima/extra/search/pqueue/AbstractQueueSearchForActions.java
@@ -1,10 +1,6 @@
 package aima.extra.search.pqueue;
 
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Queue;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Supplier;
 
 import aima.core.search.api.Node;
@@ -101,8 +97,16 @@ public abstract class AbstractQueueSearchForActions<A, S> implements QueueSearch
 		return frontier;
 	}
 
-	public boolean loopDo() {
-		return getSearchController().isExecuting();
+    public PriorityQueue<Node<A, S>> newFrontierPriority(Node<A, S> initialNode, Comparator<Node<A, S>> comparator) {
+        PriorityQueue<Node<A, S>> frontier = new PriorityQueue<>();
+        frontier.add(initialNode);
+        // Track if this is supported by the underlying implementation
+        frontierSupportsStateContainmentCheck = frontier.contains(initialNode.state());
+        return frontier;
+    }
+
+    public boolean loopDo() {
+        return getSearchController().isExecuting();
 	}
 
 	public Set<S> newExploredSet() {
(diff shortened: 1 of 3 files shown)


Manthan-R-Sheth:
@ctjoreilly  I have updated #250 .

extra/src/main/java/aima/extra/search/pqueue/AbstractQueueSearchForActions.java
```diff
@@ -129,10 +138,10 @@ public boolean containsState(Queue<Node<A, S>> frontier, S state) {
...
-		// NOTE: Not very efficient (i.e. linear in the size of the frontier)
-		// NOTE: by Java's PriorityQueue convention, nodes that compare lower
-		// (i.e. cost) have a higher priority.
-		return frontier.removeIf(n -> child.state().equals(n.state()) && getNodeComparator().compare(child, n) < 0);
-	}
+    public boolean removedNodeFromFrontierWithSameStateAndLowerPriority(Node<A, S> child, PriorityQueue<Node<A, S>> pqFrontier) {
+        // NOTE: Not very efficient (i.e. linear in the size of the frontier)
+        // NOTE: by Java's PriorityQueue convention, nodes that compare lower
+        // (i.e. cost) have a higher priority.
+        return pqFrontier.removeIf(n -> child.state().equals(n.state()));
```
ctjoreilly, line 145: This is incorrect. The method no longer does what it claims to do and it changes the semantics of the search algorithms that are dependent on this method to behave as it did before.

extra/src/main/java/aima/extra/search/pqueue/uninformed/GraphPriorityQueueSearch.java
```diff
@@ -45,19 +46,19 @@
 		// node <- a node with STATE = problem.INITIAL-STATE
 		Node<A, S> node = newRootNode(problem.initialState());
 		// frontier <- a priority queue, with node as the only element
-		Queue<Node<A, S>> frontier = newFrontier(node);
-		// explored <- an empty set
+        PriorityQueue<Node<A, S>> pqFrontier = newFrontierPriority(node, getNodeComparator());
+        // explored <- an empty set
```
ctjoreilly, line 50: There is no need to rename the variable from 'frontier' to 'pqFrontier' as it add no value and also makes it inconsistent with the pseudocode description.

extra/src/main/java/aima/extra/search/pqueue/uninformed/TreePriorityQueueSearch.java
```diff
@@ -42,16 +43,16 @@
 		// node <- a node with STATE = problem.INITIAL-STATE
 		Node<A, S> node = newRootNode(problem.initialState());
 		// frontier <- a priority queue, with node as the only element
-		Queue<Node<A, S>> frontier = newFrontier(node);
+		PriorityQueue<Node<A, S>> pqFrontier = newFrontierPriority(node, getNodeComparator());
```
ctjoreilly, line 46: This approach missed the need for the same change to be made to the GraphRLPriorityQueueSearch implementation as well.

extra/src/main/java/aima/extra/search/pqueue/uninformed/GraphPriorityQueueSearch.java
ctjoreilly: By explicitly using PriorityQueue<> you have now made the code dependent on a class not an interface, which would make it more difficult to switch out different priority queue implementations. That is why 'Queue' is used in the existing code.

ctjoreilly:
I am going to close the pull request without merging it into the main project. However, I would like to state it highlighted a usability issue in the existing AbstractQueueSearchForActions hierarchy in terms of expected default behavior. I have corrected this by adding AbstractPriorityQueueSearchForActions and re-organizing the hierarchy appropriately (see latests changes pushed up to the repository). I also, added comments to your code changes to indicate where I saw issues with your proposal. While not accepted, this was still a good submission as it highlighted a flaw in the existing implementation. Let me know what you think of the re-organized hierarchy and if it addresses the concerns you had that motivated this submission.

Thanks

Ciaran

Manthan-R-Sheth:
@ctjoreilly  
I completely understand that by my implementation, code dependent on class was made, but I wanted to preserve the hierarchy. This led to the class dependence. By re-organizing the hierarchy, my goal of having underlying priority queue for searches ```TreePriorityQueueSearch```, ```GraphRLPriorityQueueSearch``` and ```GraphRLPriorityQueueSearch```  is met.
Much appreciated. 
Thank you. :smile:

PR #261, opened 2017-03-12
Print 2D map

globalworming:
because of https://github.com/aimacode/aima-java/pull/254#discussion_r105550286

> It is located in aima.core.environment.map2d

ok

> Include a main(String[]) and updates inner methods to support specifying where the output should be written.

ok

> Is generalized to printing an Map3D and renamed to PrintMap2D

nope.
nope nope nope :)
rendering 3D stuff is complicated.

If you have a specific case where you need a .png for a three dimensional state, I may figure out a way to picture it. But a generic "print 3D map" is a huge deal and I bet there are lots of open source solutions for that problem, much better than what I would come up with.

### DIFF
--- a/core/src/main/java/aima/core/environment/map2d/PrintRomaniaMap.java
+++ b/core/src/main/java/aima/core/environment/map2d/PrintRomaniaMap.java
@@ -1,0 +1,69 @@
+package aima.core.environment.map2d;
+
+import aima.core.util.datastructure.Point2D;
+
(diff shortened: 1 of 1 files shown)


ctjoreilly:
Sorry, map3d is a typo, thanks.

ctjoreilly:
Hi,

What I intended by generalization was for the routine to work for any Map2D map, not just the simplified romania map. Do you think you could make more general?

PR #262, opened 2017-03-12
bidirectional breadth first search

globalworming:
bidirectional search with minimal feature set


globalworming:
ups, forgot the InState changes

PR #265, opened 2017-03-14
map is a parameter now. should now print all Map2Ds

globalworming:
run through your IDE with these example arguments `SimplifiedRoadMapOfPartOfRomania /tmp 2.3 500`


core/src/main/java/aima/core/environment/map2d/Print2DMap.java
```diff
@@ -0,0 +1,125 @@
...
+      System.exit(1);
+    }
+    if (args.length >= 3) {
+      zoom = Double.parseDouble(args[2]);
+    }
+    if (args.length >= 4) {
+      border = Integer.parseInt(args[3]);
+    }
+
+    final String fullQualifiedClassName = Print2DMap.class.getPackage().getName() + "." + args[0];
```
ctjoreilly, line 44: You could loosen this restriction by just insisting on a fully qualified class name as input argument. It would make the code more usable as the user would just need to ensure their own Map classes were on the classpath without having to restrict their package names (i.e. most people shouldn't be modifying the core packages).

core/src/main/java/aima/core/environment/map2d/Print2DMap.java
```diff
@@ -0,0 +1,125 @@
...
+    Class c= Class.forName(fullQualifiedClassName);
+    Map2D mapFromClassPath = (Map2D) c.newInstance();
+
+    calculateDimensions(mapFromClassPath);
+
+    int _width = ((Double) width).intValue();
+    int _height = ((Double) height).intValue();
```
ctjoreilly, line 51: Why the leading underscore in the variable names? (just curious if this is a specific convention - I've seen used for private class static attributes).

core/src/main/java/aima/core/environment/map2d/Print2DMap.java
```diff
@@ -0,0 +1,125 @@
...
+    BufferedImage image = new BufferedImage(_width, _height, BufferedImage.TYPE_INT_RGB);
+    final Graphics graphics = image.createGraphics();
+    graphics.setColor(Color.WHITE);
+    graphics.fillRect(0, 0, _width, _height);
+    draw(graphics, mapFromClassPath);
+
+    String path = args[1];
+    if (path.endsWith("/")) {
```
ctjoreilly, line 59: And what about Windows! ;) You can simply create a file object and ask if its a directory or not.

core/src/main/java/aima/core/environment/map2d/Print2DMap.java
```diff
@@ -0,0 +1,125 @@
...
+      path += "map";
+    }
+    if (!path.endsWith(".png")) {
+      path += ".png";
+    }
+    ImageIO.write(image, "png", new File(path));
+  }
+
+  private static void checkIfDirectoryIsReadable(String path) {
+    if (!new File(path).canRead()) throw new RuntimeException("can't read from " + path);
```
ctjoreilly, line 70: Ideally, even single lined conditionals should be braced, i.e. if () {}

ctjoreilly (approved):
Just some minor feedback, I plan to merge in as is.

Thanks

Ciaran

core/src/main/java/aima/core/environment/map2d/Print2DMap.java
globalworming: It's a convention where I work. To better distinguish class variables and local ones, local variables are prefixed with an underscore. Didn't do it for most of the code I write for aima, but sometimes muscle memory kicks in :)

Issue #264, opened 2017-03-13
AIMA4e improve BidirectionalSearch API to include fine grained action source details


ctjoreilly:
Update the  BidirectionalSearch API to support something like this:

interface BidirectionalActions {

  List<A> fromInitialStatePart()
  List<A> fromGoalStatePart()
  List<A> fromInitialStateToGoalState(); // What I was indicating above I was expecting as  a result
  List<A> fromGoalStateToInitialState(); // essential the reverse of the previous method.

}

as the return from SearchForActionsBidirectionallyFunction and update the tests accordingly.

Manthan-R-Sheth:
@ctjoreilly 
What is expected to be the output of ```fromInitialStatePart()``` and ```fromGoalStatePart()``` ?

ctjoreilly:
`fromInitialStatePart()` would be the actions to go from the initial state to the point it met with the backward search.
`fromGoalStatePart()` would be the actions to go from the goal state to the point it met with the forward search.

i.e. the last action of each sequence should lead you to the same state where the forward and backward searches met.

Manthan-R-Sheth:
@ctjoreilly is it good to be closed?
And with respect to the TODO list for AIMA4e can we have these changes :-
1. Add BidirectionalSearch Algorithm in the list of implemented algorithms.
2. Update the TODO list as well. 
3. For TODO of chapter 8, domain is already auto-populating while FOL parsing by the use of listeners. Please correct me if I am wrong here. If not, I would like to work on it.

ctjoreilly:
Hi,

On Wed, Mar 22, 2017 at 11:33 AM, Manthan-R-Sheth <notifications@github.com>
wrote:

> @ctjoreilly <https://github.com/ctjoreilly> is it good to be closed?
> And with respect to the TODO list for AIMA4e can we have these changes :-
>
>    1. Add BidirectionalSearch Algorithm in the list of implemented
>    algorithms.
>
>
Done. Is now in the README.md list.

>
>    1. Update the TODO list as well.
>
>
Done.


>
>    1. For TODO of chapter 8, domain is already auto-populating while FOL
>    parsing by the use of listeners. Please correct me if I am wrong here. If
>    not, I would like to work on it.
>
>
Removed, the new parser does automatically populate the domain now (the
tests also verify this).

Thanks

Ciaran

Manthan-R-Sheth:
Thanks.

PR #266, opened 2017-03-14
BidirectionalActions implemented

Manthan-R-Sheth:
This is an initial approach for implementing BidirectionalActions interface.

### DIFF
--- a/core/src/main/java/aima/core/search/api/BidirectionalActions.java
+++ b/core/src/main/java/aima/core/search/api/BidirectionalActions.java
@@ -1,0 +1,33 @@
+package aima.core.search.api;
+
+import java.util.List;
+
+/**
+ * @author manthan.
+ */
+public interface BidirectionalActions<A> {
+    /**
+     * Path from initial node to meeting node.
+     *
+     * @return List of actions that from initial state to meeting of two frontiers.
+     */
+    List<A> fromInitialStatePart();
+    /**
+     * Path from goal node to meeting node.
+     *
+     * @return List of actions that from goal state to meeting of two frontiers.
+     */
+    List<A> fromGoalStatePart();
+    /**
+     * Path from initial node to goal node.
+     *
+     * @return List of actions that from initial state to goal state.
+     */
+    List<A> fromInitialStateToGoalState();
+    /**
+     * Path from goal node to initial node.
+     *
+     * @return List of actions that from goal state to initial state.
+     */
+    List<A> fromGoalStateToInitialState();
+}
--- a/core/src/main/java/aima/core/search/basic/support/BasicBidirectionalActions.java
+++ b/core/src/main/java/aima/core/search/basic/support/BasicBidirectionalActions.java
@@ -1,0 +1,57 @@
+package aima.core.search.basic.support;
+
+import aima.core.search.api.BidirectionalActions;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * @author manthan.
+ */
+public class BasicBidirectionalActions<A> implements BidirectionalActions<A> {
+    private List<A> fromInitialStatePartList = new ArrayList<>();
+    private List<A> fromGoalStatePartList = new ArrayList<>();
+    private List<A> fromInitialStateToGoalStateList = new ArrayList<>();
+    private List<A> fromGoalStateToInitialStateList = new ArrayList<>();
+
+    public BasicBidirectionalActions(List<A> fromInitialStatePartList, List<A> fromGoalStatePartList) {
+        this.fromInitialStatePartList = fromInitialStatePartList;
+        this.fromGoalStatePartList = fromGoalStatePartList;
+    }
+
+    private void init() {
+        this.fromInitialStateToGoalStateList = new ArrayList<>();
+        this.fromGoalStateToInitialStateList = new ArrayList<>();
+    }
+
+    @Override
+    public List<A> fromInitialStatePart() {
+        return fromInitialStatePartList;
+    }
+
+    @Override
+    public List<A> fromGoalStatePart() {
+        return fromGoalStatePartList;
+    }
+
+    @Override
+    public List<A> fromInitialStateToGoalState() {
+        List<A> reversedFromGoalStatePart = new ArrayList<>(fromGoalStatePart());
+        Collections.reverse(reversedFromGoalStatePart);
+        init();
+        fromInitialStateToGoalStateList.addAll(fromInitialStatePart());
+        fromInitialStateToGoalStateList.addAll(reversedFromGoalStatePart);
+        return fromInitialStateToGoalStateList;
+    }
+
+    @Override
+    public List<A> fromGoalStateToInitialState() {
+        List<A> reversedFromInitialStatePart = new ArrayList<>(fromInitialStatePart());
+        Collections.reverse(reversedFromInitialStatePart);
+        init();
+        fromGoalStateToInitialStateList.addAll(fromGoalStatePart());
+        fromGoalStateToInitialStateList.addAll(reversedFromInitialStatePart);
+        return fromGoalStateToInitialStateList;
+    }
+
+}
--- a/core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
+++ b/core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
@@ -1,15 +1,20 @@
 package aima.core.search.basic.uninformed;
 
-import aima.core.search.api.*;
+import aima.core.search.api.SearchController;
+import aima.core.search.api.NodeFactory;
+import aima.core.search.api.BidirectionalActions;
+import aima.core.search.api.Node;
+import aima.core.search.api.Problem;
+import aima.core.search.api.SearchForActionsBidirectionallyFunction;
+import aima.core.search.basic.support.BasicBidirectionalActions;
 import aima.core.search.basic.support.BasicFrontierQueue;
 import aima.core.search.basic.support.BasicNodeFactory;
 import aima.core.search.basic.support.BasicSearchController;
-
-import java.util.Collections;
+import java.util.HashMap;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Queue;
-import java.util.Set;
-import java.util.HashSet;
 
 /**
  * Artificial Intelligence A Modern Approach (4th Edition): Figure ??, page ??.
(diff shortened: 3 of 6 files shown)


Manthan-R-Sheth:
@ctjoreilly  
This is with regards to #264. It is failing BidirectionalSearchTest as the API is now updated to support ```BidirectionalActions```. 
It is an initial approach but i suppose it gives desired results except for corner cases. By corner cases, I mean  the node that is found to be existing in other frontier is not added to the list. I can use manual addition though.
I have maintained a HashMap to get the parent of the child node in the other frontier(if it exists). ```newChildNode()``` assigns the new parent of the present frontier and so I had to use the HashMap.
Kindly give your suggestions.

Manthan-R-Sheth:
@ctjoreilly 
All functions work as expected. Review it now.

core/src/main/java/aima/core/search/api/BidirectionalActions.java
```diff
@@ -0,0 +1,13 @@
+package aima.core.search.api;
+
+import java.util.List;
+
+/**
+ * Created by manthan on 13/3/17.
+ */
```
ctjoreilly: @author manthan

core/src/main/java/aima/core/search/api/BidirectionalActions.java
```diff
@@ -0,0 +1,13 @@
...
+public interface BidirectionalActions<A> {
```
ctjoreilly: Could you add JavaDoc describing the methods.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -5,11 +5,7 @@
 import aima.core.search.basic.support.BasicNodeFactory;
 import aima.core.search.basic.support.BasicSearchController;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.Queue;
-import java.util.Set;
-import java.util.HashSet;
+import java.util.*;
```
ctjoreilly: Could we list just the classes imported and avoid .*.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -22,68 +18,87 @@
...
+            this.nextNodeToBeEvaluated = null;
+            return bidirectionalActions;
         }
         Node<A, S> revNode = newRootNode(reverseProblem.initialState(), 0);
         Queue<Node<A, S>> front = newFIFOQueue(node);
         Queue<Node<A, S>> back = newFIFOQueue(revNode);
-        Set<S> exploredFront = newExploredSet();
-        Set<S> exploredBack = newExploredSet();
+        HashMap<S, Node<A, S>> exploredFront = newExploredMap();
+        HashMap<S, Node<A, S>> exploredBack = newExploredMap();
```
ctjoreilly: These should be 'Map<..>' and not 'HashMap<...>' - i.e. avoid restricting to a specific implementation.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -22,68 +18,87 @@
...
 
-        exploredFront.add(node.state());
-        exploredBack.add(revNode.state());
+        exploredFront.put(node.state(), null);
+        exploredBack.put(revNode.state(), null);
```
ctjoreilly: newExploredMap() could take an state argument so that these line would not need to be repeated. e.g.:
newExploredMap(node.state())

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -22,68 +18,87 @@
...
         Node<A, S> solutionNode;
         while (!front.isEmpty() || !back.isEmpty()) {
 
             if (front.isEmpty() || back.isEmpty()) {
-                return failure();
+                this.previousMeetingOfTwoFrontiers = null;
```
ctjoreilly, line 57: I think you should still use a failure() method here, which implements its own interpretation of how failure is represented for this algorithm (all empty lists returned should suffice). I realize you are doing this due to implementing BidirectionalActions interface but this should be cleaned up.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -22,68 +18,87 @@
...
+                return bidirectionalActions;
             }
             // Existence of path is checked from both ends of the problem.
             solutionNode = pathExistsBidirectional(front, exploredFront, exploredBack, originalProblem);
             if (solutionNode != null) {
-                List<A> sol = solution(solutionNode);
-                Collections.reverse(sol);
-                return sol;
+                this.meetingOfTwoFrontiers = solutionNode;
```
ctjoreilly: I think you could replace these two conditional blocks with a method call something like:

if (isSolution(pathExistsBidirectional(front, exploredFront, exploredBack, originalProblem), true)) {
return bidirectionalSearch;
}

where the 2nd argument represents this.fromFront's value to assign.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -107,11 +122,65 @@
         return frontier;
     }
 
-    public Set<S> newExploredSet() {
-        return new HashSet<>();
+    public HashMap<S, Node<A, S>> newExploredMap() {
+        return new HashMap<>();
```
ctjoreilly: public Map<...

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -22,68 +18,87 @@
  * frontiers.
  * @author manthan
  */
-public class BidirectionalSearch<A, S> implements SearchForActionsBidirectionallyFunction<A, S> {
+public class BidirectionalSearch<A, S> implements SearchForActionsBidirectionallyFunction<A, S>, BidirectionalActions<A> {
```
ctjoreilly: Not a good idea having the search algorithm implement BidirectionalActions<A>. What if apply is called twice with two different arguments? Ideally, just create a simple:

aima.core.search.basic.support.BasicBidirectionalActions

class and use it.

ctjoreilly (changes requested):
Getting close, could you review feedback and make appropriate changes. The key change is not to have the search algorithm implement the BidirectionalActions interface. Once updated could you squash/rebase as necessary so that I can apply patch and try out before merging.

Thanks

Ciaran

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
Manthan-R-Sheth: @ctjoreilly, By the present implementation, failure() should be called whenever there is an empty list condition.

core/src/main/java/aima/core/search/basic/support/BasicBidirectionalActions.java
```diff
@@ -0,0 +1,86 @@
...
+    private List<A> fromInitialStatePartList = new ArrayList<>();
+    private List<A> fromGoalStatePartList = new ArrayList<>();
+    private List<A> fromInitialStateToGoalStateList = new ArrayList<>();
+    private List<A> fromGoalStateToInitialStateList = new ArrayList<>();
+    private Node<A, S> previousMeetingOfTwoFrontiers;
+    private Node<A, S> nextNodeToBeEvaluated;
+    private Node<A, S> meetingOfTwoFrontiers;
+    private boolean fromFront;
+    protected SearchController<A, S> searchController = new BasicSearchController<>();
+
```
ctjoreilly: This class is too complex for a basic support class as well as being too coupled to the current BidirectionalSearch's implementation approach. The constructor for this class should only need to take the fromInitialListPartList and the fromGoalStatePartList and to derive the other two lists from them by joining/sorting appropriately. The other logic should migrate back into BidirectionalSearch as supporting routines.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -1,15 +1,13 @@
 package aima.core.search.basic.uninformed;
 
 import aima.core.search.api.*;
```
ctjoreilly: Expand api.* to classes actually required.

ctjoreilly (changes requested):
Almost there, just a bit more refactoring and we should be good. Thanks.

core/src/main/java/aima/core/search/basic/support/BasicBidirectionalActions.java
Manthan-R-Sheth: Ya I realise that.
Am on it.

Manthan-R-Sheth:
@ctjoreilly 
Updated the PR.

Manthan-R-Sheth:
@ctjoreilly
Updated the PR and squashed the commits.

core/src/main/java/aima/core/search/basic/support/BasicBidirectionalActions.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public List<A> fromInitialStateToGoalState() {
+        List<A> sol = fromGoalStatePart();
+        Collections.reverse(sol);
+        fromInitialStateToGoalStateList.addAll(fromInitialStatePart());
+        fromInitialStateToGoalStateList.addAll(sol);
+        return fromInitialStateToGoalStateList;
+    }
```
ctjoreilly: What would be returned if this method was called twice in a row. I think it would double the size of the list of actions returned. Setting up fromInitialStateToGoalStateList and fromGoalStateToInitialStateList should only need to happen once in the constructor or a special/private init method.

core/src/main/java/aima/core/search/basic/support/BasicBidirectionalActions.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public List<A> fromGoalStatePart() {
+        return fromGoalStatePartList;
+    }
+
+    @Override
+    public List<A> fromInitialStateToGoalState() {
+        List<A> sol = fromGoalStatePart();
+        Collections.reverse(sol);
```
ctjoreilly: You've actually reversed the 'fromGoalStatePartList' corrupting for subsequent calls to fromGoalStatePart(). You need to do:

List<A> reversedFromGoalStatePart = new ArrayList<>(fromGoal...);

ctjoreilly (changes requested):
Very close but please review my comments related to BasicBidirectionalActions, address them and write a BasicBidirectionalActionsTest class demonstrating it works correctly when the API method are called several times in a row - i.e. these API calls should be idempotent.

Manthan-R-Sheth:
@ctjoreilly 
I have fixed the problem by an ```init()``` that initialises ```fromInitialStateToGoalStateList ``` and ```fromGoalStateToInitialStateList```. Supporting tests are also added.

ctjoreilly:
Can you please rebase/squash as need so I can apply as a patch to try out (currently does not work).

Manthan-R-Sheth:
@ctjoreilly 
It should work now.

PR #267, opened 2017-03-14
Added BasicPriorityFrontierQueue and Tests

manu4rhyme-zz:


ctjoreilly:
Notes from Avinash on this pull request:

"So, I have completed its implementation, as well as a test class comparing it with the original BasicFrontierQueue class, and the results look great.

There is just one minor issue with using a TreeSet and it is summarized below:

1) If the comparator used for ordering a TreeSet returns 0, then the TreeSet treats the two objects being compared as equal and it is now allowed in a Set.
2) This can be avoided very easily by modifying the comparator, such that it first checks pathCost, but if pathCosts are equal, then it doesn't simply returns 0, 
but instead further checks for equality of the objects as well. 
3) Therefore, directly using the comparator provided by NodeFactory can cause problems. 

In the test class, I've used a custom comparator instead(a rather big lambda), that adds that object check (in case path costs are equal)
In the future, we can have like, a new comparator, that wraps over the old one, and checks for object equality (when the underlying comparator returns 0). 

The results on the other hand look great and is summarized as follows:
1) For a small input it performed very similar to BasicFrontierQueue
2) For a large random input (as in the test class), BasicPriorityFrontierQueue performs almost 50 times faster while giving the same output as BasicFrontierQueue.

In any case, I believe the comparator issue is not that big of a problem, when we're getting this complexity gain. "

core/src/main/java/aima/core/search/basic/support/BasicPriorityFrontierQueue.java
```diff
@@ -0,0 +1,220 @@
...
+	//
+	// Queue
+	@Override
+	public boolean add(Node<A, S> node) {
+		
+		//iterating until there is no node with same state and lower priority
+		while(stateNodeMap.containsKey(node.state())) {
+			//find a node having the same state and lower priority (or a node that compares higher)
+			Node<A,S> toDelete = (stateNodeMap.get(node.state())).higher(node);
+			if(toDelete == null) {
```
ctjoreilly: Instead of using higher() could you use tailSet(node, false) more compactly?

test/src/test/java/aima/test/unit/search/basic/BasicPriorityFrontierQueueTest.java
```diff
@@ -0,0 +1,131 @@
...
+		
+		nodeFactory = new BasicNodeFactory<Integer, String>();
+		
+		//Custom comparator to avoid duplicates in a TreeSet
+		//Checks on pathCost first, followed by state (only for a clearer output), and finally followed by object equality
+		
+		comparator = (Node<Integer, String> node1, Node<Integer, String> node2)-> 
+					 ((node1.pathCost() == node2.pathCost()) ? ((node1.state().equals(node2.state())) ? (node1.equals(node2) ? 0 : 1) : (node1.state().compareTo(node2.state()))) 
+					 : (int)Math.signum((node1.pathCost() - node2.pathCost())));
+
```
ctjoreilly: A possibly more compact/readable version of this would be:

comparator = Comparator			
		.comparing(Node<Integer, String>::pathCost)
		.thenComparing(Comparator.comparing(Node<Integer, String>::state))
		.thenComparing((n1, n2) -> Objects.equals(n1, n2) ? 0 : 1);

test/src/test/java/aima/test/unit/search/basic/BasicPriorityFrontierQueueTest.java
```diff
@@ -0,0 +1,131 @@
...
+		nodeFactory = new BasicNodeFactory<Integer, String>();
+		
+		//Custom comparator to avoid duplicates in a TreeSet
+		//Checks on pathCost first, followed by state (only for a clearer output), and finally followed by object equality
+		
+		comparator = (Node<Integer, String> node1, Node<Integer, String> node2)-> 
+					 ((node1.pathCost() == node2.pathCost()) ? ((node1.state().equals(node2.state())) ? (node1.equals(node2) ? 0 : 1) : (node1.state().compareTo(node2.state()))) 
+					 : (int)Math.signum((node1.pathCost() - node2.pathCost())));
+		
+	    // Small Test Cases for fastPQ
```
ctjoreilly: However, the comparator and the fastQ do not work well together, odd behavior, try:

for (int i = 0; i < 10; i++) {			
 	System.out.println(i+" = "+ fastPQ.size());
	fastPQ.add(node);
}

and compare with a normal priority queue and you'll see it does not work as expected.

This needs to be rethought, as the comparator is too complex to expect a user of the API to setup correctly. An approach may be inside of BasicPriorityFrontierQueue would be to create a NodeContainer object that holds onto an insertion id (that auto increments and doesn't repeat) which is taken into account in the internal 'queue's comparator but not the 'stateNodeMap' comparator.

core/src/main/java/aima/core/search/basic/support/BasicPriorityFrontierQueue.java
```diff
@@ -0,0 +1,220 @@
...
+	//Red black tree representation for our queue
+	private TreeSet<Node<A, S>> queue;
+	//This maps states with their corresponding nodes that are present in the queue
+	private Map<S, TreeSet<Node<A, S>>> stateNodeMap;
+	
+	public BasicPriorityFrontierQueue() {
+		this(PriorityQueue::new, HashMap::new);
+	}
+	
+	public BasicPriorityFrontierQueue(Supplier<PriorityQueue<Node<A, S>>> underlyingQueueSupplier,
```
ctjoreilly: In this instance I don't think it makes sense to pass in a Supplier for a priority queue (as this object is representing one and not using it). Instead, just pass in the compartor directly.

ctjoreilly (changes requested):
Still some work to do on the comparator semantics and setup before this is usable. However, I am very encouraged by the performance comparisons and think its definitely doable. We just want to ensure the final implementation adheres to standard priority queue semantics.

core/src/main/java/aima/core/search/basic/support/BasicPriorityFrontierQueue.java
manu4rhyme-zz: Yes, i actually thought of just passing the comparator directly, but initially kept it to show similarity in usage with BasicFrontierQueue
Fixed

core/src/main/java/aima/core/search/basic/support/BasicPriorityFrontierQueue.java
manu4rhyme-zz: Moved to a separate function.
Also, Directly removing from the NavigableSet returned by tailSet() caused ConcurrentModificationException, therefore i've used .toArray() on it first.

test/src/test/java/aima/test/unit/search/basic/BasicPriorityFrontierQueueTest.java
manu4rhyme-zz: Fixed.
Now its just, 
comparator = Comparator
.comparing(Node<Integer, String>::pathCost);
as one would pass to a priority queue

test/src/test/java/aima/test/unit/search/basic/BasicPriorityFrontierQueueTest.java
manu4rhyme-zz: I tried using a NodeContainer object, but there were many issues with using it.

I've used a rather simpler approach to fix this, and the only issue remains with implementing the iterator() function, but since its overriding a Collections method and not a Queue method, I've left it unimplemented as of now. 

Changes are summarized as below:

1) The comparator issue is solved rather easily by just using 
comparator.thenComparing((n1,n2) -> Objects.equals(n1, n2) ? 0 : 1); 
over the comparator already received in the constructor.

2) Now our TreeSet only contains unique nodes. But we maintain counts of these unique nodes in a separate HashMap, to get an effect similar to a MultiTreeSet, which required changing add/remove and other methods to update this HashMap as well

Hence meeting our requirements.

I've also added tests for adding of same objects as you mentioned above, and further cleaned some code.

ctjoreilly:
Could you rebase or squash as needed your pull request so that I can apply it as a patch (currently it fails). In addition, could you move the portion of the tests from the prepare() method related to comparing BasicPriorityFrontierQueue with a PriorityQueue's  behavior into its own @Test function.  I'd like it to clearly compare with Java's priority queue and not a  BasicFrontierQueue instantiated with a PriorityQueue, in order to ensure the new logic is well behaved.

manu4rhyme-zz:
did the changes. if issue still persists, i can do a fresh PR instead.

manu4rhyme-zz:
Any updates? Should I put a new PR?

manu4rhyme-zz:
@ctjoreilly any updates? Should I put a new PR?

ctjoreilly:
Hi, I plan to review this in detail but am currently juggling several competing tasks.

Best

Ciaran

ctjoreilly:
Note: I checked quickly and am still unable to apply this as a patch, can you squash or rebase it please, see:

https://help.github.com/articles/about-pull-request-merges/

Thanks

Ciaran

manu4rhyme-zz:
@ctjoreilly I updated it again. Thanks!

norvig:
This PR tries to implement a custom priority queue. A standard implementation of PriorityQueue already exists in the java standard library and we have used it extensively throughout the project. Let's stick with that.

Issue #206, opened 2016-03-22
AIMA4e implement Fig 25.9 Monte-Carlo-Localization from AIMA3e branch


ctjoreilly:

HasinduGajanayake:
Dear All,I will try the Monte Carlo localization problem

PR #268, opened 2017-03-15
monte carlo localization interface

globalworming:
I searched the issues and found the first one pretty interesting #206
I already started working on it, but before my next pull request gets to complex, I wanted to lay some groundwork.

The algorithm in the 3rd ed book fig 25.9 is pretty complex and not very generic. The interface I propose is a generic "a Monte Carlo Localization is figuring out ones own state through percepts from the environment". This generic definition is closer to what wikipedia say to [MCL](https://en.wikipedia.org/wiki/Monte_Carlo_localization) The whole implementation with noisy sensors etc can be built on top.

If you agree with this interface, I'd like to provide an example. The AIMA3 example is pretty complex too and I am not willing to copy/implement this whole 2D geometry, ray casting etc. My example would be a much simpler continuous environment, if that's ok. If you like, I can provide a "Story" and a general description of the environment and the agent therein later.

### DIFF
--- a/core/src/main/java/aima/core/agent/api/Belief.java
+++ b/core/src/main/java/aima/core/agent/api/Belief.java
@@ -1,0 +1,13 @@
+package aima.core.agent.api;
+
+/**
+ * A belief as a state with a probability
+ * @author wormi
+ */
+public interface Belief<S> {
+
+  double getWeight();
+
+  S getState();
+
+}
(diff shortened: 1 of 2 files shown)


ctjoreilly:
I need to think about this some more but am curious where this will lead. As your proposal diverges from the description from the book for now can you work on this in the 'extra' module (i.e. move the two initial interfaces that you are proposing to live under there). Also, as none of the supporting probability code is available in AIMA4e currently I am not sure of how much of an impact this will be on what you are proposing.

globalworming:
ok, will move it to extra

PR #269, opened 2017-03-16
IterativeDeepeningAlphaBetaSearch implemented

Manthan-R-Sheth:
Most of the adaptation is from AIMA3e branch. 

**Changes Made** :-
1. ```ActionStore<A>``` inner class is replaced with a ```TreeSet``` to improve the complexity of insertion.
1. Code is cleaned and adjusted according to AIMA4e conventions.

**Path Ahead** :-
1. An interface will be made to increase the functionality, or ```SearchForAdversarialActionFunction``` can be used directly.
1. Tests will be written. 

Since AphaBetaSearch is used with an additional ```depth``` parameter, a class similar to ```DepthLimitedTreeSearch``` could have been used.

### DIFF
--- a/extra/src/main/java/aima/extra/search/uninformed/IterativeDeepeningAlphaBetaSearch.java
+++ b/extra/src/main/java/aima/extra/search/uninformed/IterativeDeepeningAlphaBetaSearch.java
@@ -1,0 +1,133 @@
+package aima.extra.search.uninformed;
+
+import aima.core.search.api.Game;
+import aima.core.search.api.SearchForAdversarialActionFunction;
+
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+
+/**
+ * Implements an iterative deepening Minimax search with alpha-beta pruning and
+ * action ordering. Maximal computation time is specified in seconds. The
+ * algorithm is implemented as template method and can be configured and tuned
+ * by subclassing.
+ *
+ * @author Ruediger Lunde
+ * @author manthan.
+ * 
+ * @param <S>
+ *            Type which is used for states in the game.
+ * @param <A>
+ *            Type which is used for actions in the game.
+ * @param <P>
+ *            Type which is used for players in the game.
+ */
+public class IterativeDeepeningAlphaBetaSearch<S, A, P> implements SearchForAdversarialActionFunction<S, A> {
+    private Game<S, A, P> game;
+    private double utilMax;
+    private double utilMin;
+    private int currDepthLimit;
+    private boolean heuristicEvaluationUsed = true;
+    private Timer timer;
+
+    public static <S, A, P> IterativeDeepeningAlphaBetaSearch<S, A, P> createFor(Game<S, A, P> game, double utilMin, double utilMax, int time) {
+        return new IterativeDeepeningAlphaBetaSearch<>(game, utilMin, utilMax, time);
+    }
+
+    public IterativeDeepeningAlphaBetaSearch(Game<S, A, P> game, double utilMin, double utilMax, int time) {
+        this.game = game;
+        this.utilMin = utilMin;
+        this.utilMax = utilMax;
+        this.timer = new Timer(time);
+    }
+
+    @Override
+    public A apply(S state) {
+        P player = game.player(state);
+        List<A> results = game.actions(state);
+        timer.start();
+        Map<Double, A> newResults;
+        for (currDepthLimit = 2; !timer.timeOutOccured() && heuristicEvaluationUsed; currDepthLimit++) {
+            heuristicEvaluationUsed = false;
+            newResults = new TreeMap<>();
+            for (A action : results) {
+                double value = minValue(game.result(state, action), player, Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY, 1);
+                if (timer.timeOutOccured())
+                    break;
+                newResults.put(value, action);
+            }
+            if (newResults.size() > 0) {
+                results = newResults.values().stream().collect(Collectors.toList());
+                if (!timer.timeOutOccured()) {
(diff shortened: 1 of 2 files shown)


extra/src/main/java/aima/extra/search/uninformed/IterativeDeepeningAlphaBetaSearch.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+
+/**
+ * @author manthan.
+ */
+public class IterativeDeepeningAlphaBetaSearch<A, S, P> implements SearchForAdversarialActionFunction<S, A> {
```
ctjoreilly: Could you order <A, S, P> the same as AlphaBetaSearch i.e. <S, A, P> and add @param descriptions to the class JavaDoc in a similar way.

ctjoreilly (changes requested):
This looks good, could you address the minor comment I made and also rebase/squash as necessary so I can apply as a patch and try out before merging.

Thanks

Ciaran

Manthan-R-Sheth:
@ctjoreilly 
Updated as said.

extra/src/main/java/aima/extra/search/uninformed/IterativeDeepeningAlphaBetaSearch.java
```diff
@@ -0,0 +1,126 @@
...
+import aima.core.search.api.Game;
+import aima.core.search.api.SearchForAdversarialActionFunction;
+
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+
+/**
+ * @author manthan.
```
ctjoreilly: As this is based largely on the AIMA3e version of the code could you add the original author's name here as well (i.e. @author Ruediger Lunde, I would also suggest taking over the description). The convention is that the author who contributed the most to a particular implementation should be listed first. If an author has been involved or code is closely based on a prior implementation, there name should remain included.

test/src/test/java/aima/test/unit/search/adversarial/SearchForAdversarialActionFunctionTest.java
```diff
@@ -28,22 +29,23 @@
...
+	public <S, A, P> A searchForAdversarialAction(Game<S, A, P> game, Integer... params) {
+		SearchForAdversarialActionFunction<S, A> searchFn;
+		if (params != null) {
+			if ("MinimaxDecision".equals(searchFunctionName)) {
+				searchFn = new MinimaxDecision<>(game);
+			} else {
+				searchFn = new AlphaBetaSearch<>(game);
+			}
+		} else {
+			searchFn = new IterativeDeepeningAlphaBetaSearch<>(game, params[0], params[1], params[2]);
```
ctjoreilly: NOTE: IterativeDeepeningAlphaBetaSearch is never tested! At this point params is null and the way the test below is configured AlphaBetaSearch is all that is called and only one test contains the params for the IterativeDeepening calls. This code should be re-configured to just be:

	public <S, A, P> A searchForAdversarialAction(Game<S, A, P> game, Integer... params) {
		SearchForAdversarialActionFunction<S, A> searchFn;
		
		if ("MinimaxDecision".equals(searchFunctionName)) {
			searchFn = new MinimaxDecision<>(game);
		} else if ("AlphaBetaSearch".equals(searchFunctionName)) {
			searchFn = new AlphaBetaSearch<>(game);
		} else {
			searchFn = new IterativeDeepeningAlphaBetaSearch<>(game, params[0], params[1], params[2]);
		}
		
		return searchFn.apply(game.initialState());
	}

test/src/test/java/aima/test/unit/search/adversarial/SearchForAdversarialActionFunctionTest.java
```diff
@@ -111,7 +113,14 @@ public void testAIMA3eFig5_2() {
...
-		
+
 		Assert.assertEquals(new GoAction("B"), searchForAdversarialAction(game));
+
+		Game<InState, GoAction, String> game2 = new BasicGame<>(() -> new InState("A"), aima3eFig5_2PlayerFn,
+				Map2DFunctionFactory.getActionsFunction(aima3eFig5_2),
+				Map2DFunctionFactory.getResultFunction(aima3eFig5_2), aima3eFig5_2TerminalTestPredicate,
+				aima3eFig5_2UtilityFn);
+		Assert.assertEquals(new GoAction("B"), searchForAdversarialAction(game2, 100, 0, 200));
+
```
ctjoreilly: This is just repeating the prior test and adding params (not used currently) specifically for the iterative deepening search. All you need to do is add the params to each of the searchForAdversarialAction calls, as they are only intended to be used by the IterativeDeepening algorithm currently and don't affect the other 2 algorithms. All it needs to be is this:

	@Test
	public void testAIMA3eFig5_2() {
		Game<InState, GoAction, String> game = new BasicGame<>(() -> new InState("A"), aima3eFig5_2PlayerFn,
				Map2DFunctionFactory.getActionsFunction(aima3eFig5_2),
				Map2DFunctionFactory.getResultFunction(aima3eFig5_2), aima3eFig5_2TerminalTestPredicate,
				aima3eFig5_2UtilityFn);
		Assert.assertEquals(new GoAction("B"), searchForAdversarialAction(game, 100, 0, 200));

	}

ctjoreilly (changes requested):
Please rebase/squash as before when you re-submit.

Thanks

Ciaran

Manthan-R-Sheth:
@ctjoreilly 
Ya i realise the mistake of ```IterativeDeepeningAlphaBetaSearch``` never being tested.
Have fixed it though.

Issue #217, opened 2016-04-13
Add Implementation of Air-Cargo-problem Fig 10.1 for AIMA3e branch


smsubham:
No implementation of this was found so added this issue.

smsubham:
Have opened this issue and will be submitting a patch for it soon. @ctjoreilly

anurag-rai:
Hi @smsubham , I'm not sure if this (like #218) requires an implementation... These cases can be modeled as Test Cases, perhaps, for (say) Search algorithms. Although, if this is an issue altogether would depend on @ctjoreilly

smsubham:
Hi @anurag-rai
Good to hear from you. Am aware of this.
Actually yesterday i had an opportunity to have a look at [aima-python](https://github.com/aimacode/aima-python) page , found these fig no's (issues #217 #218 #219 #220 #221 ) in missing list of algorithms which are yet to be implemented in AIMA-Python. Have tagged @ctjoreilly sir in these issue comments , so that he can have a look at these issues and clarify if its required. But i think these issues should be accepted , most of these will be modeled as test cases though.

anurag-rai:
Oh ok. Good to know the reference!

divsidhu-26:
@ctjoreilly i would like to work on this isssue if it is still open.......

ctjoreilly:
Yes.

divsidhu-26:
@ctjoreilly can u give a little description how you want this code to be implemented......
Do i need to build a new parser or can i use the FOLParser directly?

ctjoreilly:
You can likely get away with the FOL Parser if you use a simple pre-parser to break up the file into the specific sections (i.e. Action, PRECOND, and Effect sections). The FOL parser on the aima4e branch in the 'extra' module is better to use as it doesn;t require that you specify an FOLDomain manually beforehand (i.e. it automatically extracts predicates, constants and functions based on the parse).

Best

Ciaran

samagra14:
This has already been implemented. Therefore closing this issue.

PR #272, opened 2017-03-19
Issue #217 : Added Implementation of Air-Cargo-problem Fig 10.1 for AIMA3e branch

divsidhu-26:
Please check if the solution is correct. Also I wasnt sure if i have to write comments to explain the code. Please update me on that and also tell any changed in the code if they were to be made. Isssue #217

### DIFF
--- a/aima-core/src/main/java/aima/core/learning/knowledge/Classical_Planning/AirCargoProblem.java
+++ b/aima-core/src/main/java/aima/core/learning/knowledge/Classical_Planning/AirCargoProblem.java
@@ -1,0 +1,117 @@
+package aima.core.learning.knowledge.Classical_Planning;
+
+import java.io.*;
+import java.util.*;
+public class AirCargoProblem {
+	public static Cargo Load(Cargo c,Plane p,Airport a){
+		if(!(at(c,a) && at(p,a)))
+			return null;
+		notAt(c,a);
+		In(c,p);
+		return c;
+	}       
+    
+	public static Cargo Unload(Cargo c,Plane p,Airport a){
+		if(!(in(c,p) && at(p,a)))
+			return null;
+		At(c,a);
+		notIn(c,p);
+		return c;
+	}
+    
+	public static Plane Fly(Plane p,Airport from,Airport to){
+		if(!at(p,from))
+			return null;
+		notAt(p,from);
+		At(p,to);
+		return p;
+	}
+    
+
+	public static void At(Plane p,Airport a){
+		p.airport = a;
+	}
+	public static void At(Cargo c,Airport a){
+		c.loaded = false;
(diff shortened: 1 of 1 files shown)


aima-core/src/main/java/aima/core/learning/knowledge/Classical_Planning/AirCargoProblem.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+class Plane{
+	String name;
+	Airport airport;
+	Plane(String p){
+		this.name = p;
+	}
+	public boolean equals(Plane a){
+    	if((this.name).equals(a.name) && (this.airport).equals(a.airport)){
```
duchuyctlk: How about directly returning the result? 😄 

```java
return this.name.equals(a.name) && this.airport.equals(a.airport);
```

aima-core/src/main/java/aima/core/learning/knowledge/Classical_Planning/AirCargoProblem.java
```diff
@@ -0,0 +1,120 @@
...
+		System.out.println(at(C1,JFK)+" "+at(C2,SFO));
+	}
+}
+class Airport{
+	String name;
+    Airport(String s){
+    	this.name = s;
+    }
+    public boolean equals(Airport a){
+    	if((this.name).equals(a.name)){
```
duchuyctlk:
```java
return this.name.equals(a.name);
```

😄

aima-core/src/main/java/aima/core/learning/knowledge/Classical_Planning/AirCargoProblem.java
divsidhu-26: yea sure and do i need to comment explanations along with the code

duchuyctlk: Oh, this approach is really clear already, i don't think that explanation required here.

divsidhu-26: ok i will erase the comments............

ctjoreilly:
This is a neat little example. Unfortunately, it takes an imperative programming approach to what is intended to be a declarative approach (i.e. PDDL). The ideal way to tackle this problem is to have a specialized parser based on the FOL one, that knows how to extract out the Action, PRECOND, and Effect sections - validate they are of the correct form and make them available to the caller. The Air Cargo example itself then would just be a text file mirroring Figure 10.1 exactly.

divsidhu-26:
ok i will try to do it with that approach..........

PR #273, opened 2017-03-20
Added guava library to write shorter and simpler code for string manipulation

ngrj93:
**Updates**

- Added guava-19.0.jar in the aima-core/lib/guava directory 
- Updated pom.xml
- Updated build.xml


ctjoreilly:
Hi,

I'm afraid one of the goals of AIMA3e and AIMA4e is to have 0 dependencies from the 'core' modules on third party libraries (excluding testing support via JUnit).

PR #274, opened 2017-03-20
Shortened code for describeAttributes() in ObjectWithDynamicAttributes.java

ngrj93:
**Updates**

- Updated describeAttributes() to a simpler form


ctjoreilly:
As mentioned in other pull request this introduces an unwanted dependency on a third party library to a 'core' module. An alternative would be to leverage Java 8's StringJoiner class to simplify this code:

AIMA3e as well as AIMA4e lowest JDK support is 8 now as well.

PR #275, opened 2017-03-21
Initial Tidy up done on RecursiveBestFirstSearch

manu4rhyme-zz:
@ctjoreilly Added helper functions and made SuccessorNode implement Node (for better use cases of node.state() instead of node.n.state() etc.)

### DIFF
--- a/core/src/main/java/aima/core/search/basic/informed/RecursiveBestFirstSearch.java
+++ b/core/src/main/java/aima/core/search/basic/informed/RecursiveBestFirstSearch.java
@@ -52,7 +52,7 @@ public class RecursiveBestFirstSearch<A, S> implements SearchForActionsFunction<
 	@Override
 	public List<A> apply(Problem<A, S> problem) {
 		// return RBFS(problem, MAKE-NODE(problem.INITIAL-STATE), infinity)
-		return rbfs(problem, new SuccessorNode(nodeFactory.newRootNode(problem.initialState()), this::h),
+		return rbfs(problem, newRootNode(problem.initialState(), 0),
 				Double.POSITIVE_INFINITY).result();
 	}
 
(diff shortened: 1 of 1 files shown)


manu4rhyme-zz:
@ctjoreilly 
Also the code line 

```
`// alternative <- the second-lowest f-value among successors

double alternative = successors.size() > 1 ? successors.get(1).f : best.f;`
```

How would the value at index 1 be always have the second-lowest f-value among the successors ?

ctjoreilly:
It looks like I introduced a defect in my prior commit where I replaced the lines:

Collections.sort(successors, (s1, s2) -> Double.compare(s1.f, s2.f));
SuccessorNode best = successors.get(0);

with: 
SuccessorNode best = successors.stream().min(Comparator.comparingDouble(s -> s.f)).get();

I would switch it back to something equivalent like:

Collections.sort(successors, Comparator.comparingDouble(s -> s.f));
SuccessorNode best = successors.get(0);

Good catch,

Thanks

manu4rhyme-zz:
@ctjoreilly 
We can also write helper functions for this, to maintain similarity with Pseudo-code. Also since we only need the smallest two values, we can do a linear O(n)  search inside these helper functions instead of a full sort, which would also make it more efficient for larger test cases (Not particularly necessary here, but is something that can be done)

ctjoreilly:
Sounds like a good strategy. Could you make these changes first and I'll consider all together in a squashed pull request.

Thanks

Ciaran

manu4rhyme-zz:
@ctjoreilly 
Fixed

core/src/main/java/aima/core/search/basic/informed/RecursiveBestFirstSearch.java
```diff
@@ -52,27 +52,27 @@
...
 				Double.POSITIVE_INFINITY).result();
 	}
 
 	// function RBFS(problem, node, f_limit) returns a solution, or failure and
 	// a new f-cost limit
 	public Result rbfs(Problem<A, S> problem, SuccessorNode node, double f_limit) {
 		// if problem.GOAL-TEST(node.STATE) then return SOLUTION(node)
-		if (searchController.isGoalState(node.n, problem)) {
-			return new Result(searchController.solution(node.n));
+		if (problem.isGoalState(node.state())) {
```
ctjoreilly, line 63: You want to add a 'isGoalState()' helper function call here, that actually calls the searchController like before i.e.:

searchController.isGoalState(node.n, problem)

and not the problem directly as you have done. The reason for this is it lets someone override the behavior in the search controller, for e.g. if they want to visit more than 1 goal state. The reason for passing the Node to the searchController function - problem.isGoalState only gets to check the state and not necessarily the node information associated with it.

core/src/main/java/aima/core/search/basic/informed/RecursiveBestFirstSearch.java
```diff
@@ -122,6 +122,48 @@ public double h(Node<A, S> node) {
...
+	public SuccessorNode newChildNode(Problem<A, S> problem, SuccessorNode node, A action) {
+		return new SuccessorNode(nodeFactory.newChildNode(problem, node.n, action), this::h);
+	}
+	
+	public SuccessorNode getLowestFValueNode(List<SuccessorNode> nodes) {
+		return nodes.stream().min(Comparator.comparingDouble(s -> s.f)).get();
+	}
+
+	public double getSecondLowestFValue(List<SuccessorNode> nodes) {
+
```
ctjoreilly, line 146: you may be able to simplify this logic if you passed the 'best' value as an argument to this function.

ctjoreilly (approved):
This is a good improvement. I am going to merge your changes as they stand but would like you to address the additional couple of comments I added.

Thanks

Ciaran

core/src/main/java/aima/core/search/basic/informed/RecursiveBestFirstSearch.java
manu4rhyme-zz: True, I'll fix it. 
Also, there are still many algorithms including BreadthFirstSearch, that contain problem.isGoalState() and not searchController.isGoalState(), shouldn't they be fixed as well , or if I'm mistaken, there is a reason to keep them ?

ctjoreilly: Yes they should, I think it would be good to have all the core algorithms implemented in a consistent way.

PR #277, opened 2017-03-25
Optimise UniformCostSearch algorithm.

Manthan-R-Sheth:
The aim of the PR is to optimise UniformCostSearch algorithm.
1. Linear time check for state containment is replaced by the built in queue containment and ```BasicFrontierQueue``` is used with underlying Priority queue implementation and hasmap.
1. Use of ```AbstractPriorityQueueSearchForActions``` and ```QueueSearchForActions``` would have been better but i refrained from doing so to prevent circular dependence between core and extra modules.
1. I have written a separate test class, it can be merged with other test class as well.

### DIFF
--- a/core/src/main/java/aima/core/search/basic/uninformed/UniformCostSearch.java
+++ b/core/src/main/java/aima/core/search/basic/uninformed/UniformCostSearch.java
@@ -1,6 +1,12 @@
 package aima.core.search.basic.uninformed;
 
-import java.util.*;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


ctjoreilly:
I've considered these optimizations but would prefer not to add them to the core implementation. The reason is the alternative implementation of this and related 'informed' algorithms all perform these optimizations in the 'extra' module. I would like the basic versions in the 'core' module to highlight, as the do, the drawbacks with a rudimentary implementation that adheres to the pseudo-code strongly. However, if you could submit your test case in a separate pull request for uniform cost search I think that would be a useful addition.

Thanks

Ciaran

PR #278, opened 2017-03-27
Add test for UniformCostSearch

Manthan-R-Sheth:
UniformCostSearch is tested as in #277

### DIFF
--- a/test/src/test/java/aima/test/unit/search/uninformed/UniformCostSearchTest.java
+++ b/test/src/test/java/aima/test/unit/search/uninformed/UniformCostSearchTest.java
@@ -1,0 +1,47 @@
+package aima.test.unit.search.uninformed;
+
+import aima.core.environment.map2d.GoAction;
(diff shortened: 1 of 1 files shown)


ctjoreilly:
I have manually merged this pull request. I added three additional versions of UniformCostSearch to the test. However, 1 of them is failing, i.e.:

"extra.UniformCostQueueSearch.graphrl"

Could you investigate (it will be a good exercise figuring out what is going wrong with this particular implementation). See:

https://github.com/aimacode/aima-java/issues/283

Thanks

Ciaran

Manthan-R-Sheth:
@ctjoreilly 
Ya I am on it .
Thanks.

Manthan-R-Sheth:
@ctjoreilly 
The problem is that, a node is never added to ```explored``` and its containment is checked in the ```while``` condition. Hence, being always true (```!explored.contains(node.state())```), all nodes from ```frontier``` are removed till the queue becomes empty and returns ```failure()```.
We need to find a method to populated ```explored``` and then execute the do...while loop.

PR #287, opened 2017-03-27
Fixed #286 Empty Catch Blocks

### LINKED ISSUE #286: Empty Catch Blocks
In some of the files catch blocks do not perform any action. A simple call to printStackTrace() method can be added in such blocks so that the throwable and its backtrace are printed on the standard error stream.

rajatthakur:
I have added a call to printStackTrace in each and every empty catch block in the codebase.

Thanks :smile:


ctjoreilly:
In general I agree with this but will need to run through the related code to ensure it hasn't introduced any unwanted output (i.e. the exceptions are being used like conditionals and stack traces are not really wanted).

PR #290, opened 2017-03-29
Update BidirectionalSearch.java

kanishk1225:

PR #291, opened 2017-03-29
Update BidirectionalSearch.java

kanishk1225:


ctjoreilly:
@Manthan-R-Sheth  can you review this.

ctjoreilly:
@Manthan-R-Sheth  can you review this.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -63,7 +63,7 @@
                 bidirectionalActions = new BasicBidirectionalActions<>(fromInitialStatePart(), fromGoalStatePart());
                 return bidirectionalActions;
             }
-            if (isSolution(pathExistsBidirectional(back, exploredBack, exploredFront, originalProblem), false)) {
+            if (isSolution(pathExistsBidirectional(back, exploredBack, exploredFront, reverseProblem), false)) {
```
Manthan-R-Sheth, line 66: Though ```originalProblem``` would have worked (just like it already is) because the same graph is used for both the problems (original and reverse) with only the initial state differing, ```reverseProblem```improves readability and so is fine for the purpose.

Manthan-R-Sheth:
@kanishk1225, if you could close #290 and amend the changes in this PR or add the changes and then squash it. This is to have a single PR aimed towards a particular task.

kanishk1225:
Squashed changes with #290 closing this one

PR #294, opened 2017-03-31
refactored BiDirectionalSearchTest for readability

globalworming:
Hi again. As MCL is a bit tricky, I wanted to go back to BiDirectionalSearch. Test was refactored for readability. Nothing has changed.

I want to use this pull request to point out some problems with the current implementation.

### DIFF
--- a/test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
+++ b/test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
@@ -8,13 +8,24 @@ import aima.core.search.api.BidirectionalActions;
 import aima.core.search.api.Problem;
 import aima.core.search.basic.uninformed.BidirectionalSearch;
 import aima.core.util.datastructure.Pair;
-import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.ARAD;
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.BUCHAREST;
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.FAGARAS;
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.ORADEA;
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.SIBIU;
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.TIMISOARA;
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.URZICENI;
+import static aima.core.environment.map2d.SimplifiedRoadMapOfPartOfRomania.ZERIND;
+import static org.junit.Assert.assertEquals;
 
 /**
  * @author manthan.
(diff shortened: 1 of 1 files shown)


test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
```diff
@@ -29,50 +40,57 @@
...
-    public <A, S> BidirectionalActions<A> searchForActions(Problem<A, S> originalProblem, Problem<A, S> reverseProblem) {
-        BidirectionalSearch<A, S> searchForActionsFunction = new BidirectionalSearch<A, S>();
-        return searchForActionsFunction.apply(originalProblem, reverseProblem);
+    public <A, S> BidirectionalActions<A> searchForActions(Pair<Problem<A, S>, Problem<A, S>> pair) {
+        BidirectionalSearch<A, S> search = new BidirectionalSearch<>();
+        return search.apply(pair.getFirst(), pair.getSecond());
+    }
+
+    @Test
+    public void aradToArad() throws Exception {
```
globalworming, line 49: this test is fine

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
```diff
@@ -29,50 +40,57 @@
...
-                searchForActions(problemPairtest5.getFirst(), problemPairtest5.getSecond()).fromGoalStateToInitialState());
+    public void timisoaraToUrziceni() throws Exception {
+      //Result will contain all intermediate nodes excluding initial and final nodes.
+        Pair<Problem<GoAction, InState>, Problem<GoAction, InState>> pair =
+            ProblemFactory.getSimpleBidirectionalSearchProblem(TIMISOARA, URZICENI);
+        final BidirectionalActions<GoAction> solution = searchForActions(pair);
+
+        assertEquals(Arrays.asList(
+            new GoAction(ARAD),new GoAction(SIBIU), new GoAction(FAGARAS)),
+            solution.fromInitialStatePart());
```
globalworming, line 65: here I'd like to ask: is it really important to get the different "parts" of the solution? I search, I want a solution, why bloat the interface?

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
```diff
@@ -29,50 +40,57 @@
...
-        Pair<Problem<GoAction, InState>, Problem<GoAction, InState>> problemPairtest5 = ProblemFactory.getSimpleBidirectionalSearchProblem(
-                SimplifiedRoadMapOfPartOfRomania.TIMISOARA, SimplifiedRoadMapOfPartOfRomania.URZICENI);
-        Assert.assertEquals(
-                Arrays.asList(new GoAction(SimplifiedRoadMapOfPartOfRomania.BUCHAREST),
-                        new GoAction(SimplifiedRoadMapOfPartOfRomania.FAGARAS),
-                        new GoAction(SimplifiedRoadMapOfPartOfRomania.SIBIU),
-                        new GoAction(SimplifiedRoadMapOfPartOfRomania.ARAD)),
-                searchForActions(problemPairtest5.getFirst(), problemPairtest5.getSecond()).fromGoalStateToInitialState());
+    public void timisoaraToUrziceni() throws Exception {
+      //Result will contain all intermediate nodes excluding initial and final nodes.
```
globalworming, line 58: > //Result will contain all intermediate nodes excluding initial and final nodes.

this is plain wrong. there are no nodes, the result is a list of actions which - applied to the inital state - lead to the goalstate. Excluding the last goAction makes no sense

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
```diff
@@ -29,50 +40,57 @@
...
+
+        assertEquals(Arrays.asList(
+            new GoAction(BUCHAREST),
+            new GoAction(FAGARAS),
+            new GoAction(SIBIU),
+            new GoAction(ARAD)),
+            solution.fromGoalStateToInitialState());
     }
 
+    @Test @Ignore("fails, actual is [null, Go(Oradea), null]")
```
globalworming, line 85: I added another testcase and the result is nonsensical, that's why it is ignored for now

globalworming (commented):
The interface is bloated, the results are wrong. If you and @Manthan-R-Sheth acknowledge that, I will fix the mentioned problems in the next pull request.

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
```diff
@@ -29,50 +40,57 @@
     @Parameterized.Parameter
     public String searchFunctionName;
 
-    public <A, S> BidirectionalActions<A> searchForActions(Problem<A, S> originalProblem, Problem<A, S> reverseProblem) {
-        BidirectionalSearch<A, S> searchForActionsFunction = new BidirectionalSearch<A, S>();
-        return searchForActionsFunction.apply(originalProblem, reverseProblem);
+    public <A, S> BidirectionalActions<A> searchForActions(Pair<Problem<A, S>, Problem<A, S>> pair) {
```
globalworming, line 43: this may seem picky, but what are BidirectionalActions? Actions that work in two directions? Does this name fit the terminology in the book?

> public interface BidirectionalActions<A> {
    /**
     * Path from initial node to meeting node.
     *
     * @return List of actions that from initial state to meeting of two frontiers.

is this an action? what do actions know of meeting nodes? This should be renamed as it is confusing.

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
Manthan-R-Sheth: It is a corner case that needs to be handled. It will make sense if you see ```BasicBidirectionalActions```. As far as the tests are concerned, I will make a PR wherein this corner case will be handled.

Manthan-R-Sheth:
Primary purpose of ```BidirectionalActions``` is to enhance the functionality of the search algorithm. It adds functionality to the API.

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
globalworming: https://github.com/aimacode/aima-java/pull/294#issuecomment-290656910
> Primary purpose of BidirectionalActions is to enhance the functionality of the search algorithm. It adds functionality to the API.

The name should convey it's purpose though http://www.itiseezee.com/?p=83
everything else makes it harder for everyone trying to understand the code

test/src/test/java/aima/test/unit/search/uninformed/BidirectionalSearchTest.java
globalworming: how is this a corner case? We have a simple environment, we have a simple algorithm. Why can't the implementation solve all possible problems in this environment? Why wasn't there a unit test to begin with? Why is there an incomplete implementation on the AIMA4 master branch? 

This is more a critique at @ctjoreilly for not demanding more tests..

Manthan-R-Sheth: It is a corner case because the resulting list will have ```null``` only if the meeting node is just after the initial node(state) or before the goal node(state). Hence this case needs to be handled explicitly which can be done by just checking for null condition in ```BasicBidirectionalActions``` implementation. 
And moreover, the algorithm always stands a chance of modification once the algorithm in the latest edition is released.
I will update it, though.

ctjoreilly:
I have merged this manually (minor edits). @globalworming I will try to address your concerns about the BidirectionalActions API bloat in your related pull request. Please note, that at the moment the goal is to try and stabilize the API to bidirectional search. The guts are going to have to be completely rewritten once the pseudocode for BidrectionalSearch is in place, see:

https://github.com/aimacode/aima-pseudocode/issues/41

Perhaps submitting a draft version of the pseudocode to that issue would be a good next step before doing any more major work on the current implementation as we know its going to change drastically.

PR #296, opened 2017-03-31
Using some Java8 features on DecisionTreeLearner and DataSet

AdrianBZG:
Just a few changes, using Java8 features:

- Check allExampleHaveSameClassification: DecisionTreeLearner.java (L139-L143) and chooseAttribute (L122-L137) on commit f20a270

- Check calculateGainFor on DataSet.java (L101-114)

### DIFF
--- a/aima-core/src/main/java/aima/core/learning/framework/DataSet.java
+++ b/aima-core/src/main/java/aima/core/learning/framework/DataSet.java
@@ -47,9 +47,9 @@ public class DataSet {
 		// and add them to the DataSet (ds) to return it afterwards
 		examples.stream().filter(example -> {
 			if (e.equals(example)) {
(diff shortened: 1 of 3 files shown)


ctjoreilly:
Hi,

The AIMA4e branch is the intended target for introducing JDK 8 features. While AIMA3e is now based on JDK 8 as well these types of changes to its source base are unnecessary. Also, switching to streams() for the sake of using streams() is not necessarily a good idea. See the discussion here:

https://blog.jooq.org/2015/12/08/3-reasons-why-you-shouldnt-replace-your-for-loops-by-stream-foreach/

If you can just fix the glitch and test case in a clean/separate pull request I'll re-review.

Best

Ciaran

PR #297, opened 2017-03-31
Fix bidirectional search

globalworming:
smaller interface
more unittests
unittests even work in corner cases @Manthan-R-Sheth

### DIFF
--- a/core/src/main/java/aima/core/search/api/SearchForActionsBidirectionallyFunction.java
+++ b/core/src/main/java/aima/core/search/api/SearchForActionsBidirectionallyFunction.java
@@ -1,12 +1,14 @@
 package aima.core.search.api;
 
-import java.util.function.BiFunction;
+import aima.core.util.datastructure.Pair;
+
+import java.util.List;
+import java.util.function.Function;
 
 /**
  * @author manthan
  */
 @FunctionalInterface
-public interface SearchForActionsBidirectionallyFunction<A, S> extends BiFunction<Problem<A, S>, Problem<A, S>, BidirectionalActions<A>> {
-    @Override
-    BidirectionalActions<A> apply(Problem<A, S> originalProblem, Problem<A, S> reverseProblem);
+public interface SearchForActionsBidirectionallyFunction<A, S> extends Function<Pair<Problem<A,
+    S>, Problem<A, S>>, List<A>> {
 }
--- a/core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
+++ b/core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
@@ -2,19 +2,20 @@ package aima.core.search.basic.uninformed;
 
 import aima.core.search.api.SearchController;
 import aima.core.search.api.NodeFactory;
-import aima.core.search.api.BidirectionalActions;
 import aima.core.search.api.Node;
 import aima.core.search.api.Problem;
 import aima.core.search.api.SearchForActionsBidirectionallyFunction;
-import aima.core.search.basic.support.BasicBidirectionalActions;
 import aima.core.search.basic.support.BasicFrontierQueue;
 import aima.core.search.basic.support.BasicNodeFactory;
 import aima.core.search.basic.support.BasicSearchController;
-import java.util.HashMap;
+import aima.core.util.datastructure.Pair;
+
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Queue;
+import java.util.Set;
 
 /**
  * Artificial Intelligence A Modern Approach (4th Edition): Figure ??, page ??.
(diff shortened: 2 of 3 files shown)


core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -28,80 +29,120 @@
  * @author manthan
  */
 public class BidirectionalSearch<A, S> implements SearchForActionsBidirectionallyFunction<A, S> {
-    private BidirectionalActions<A> bidirectionalActions;
-    private List<A> fromInitialStatePartList = new ArrayList<>();
-    private List<A> fromGoalStatePartList = new ArrayList<>();
-    private Node<A, S> previousMeetingOfTwoFrontiers;
-    private Node<A, S> meetingOfTwoFrontiers;
-    private Node<A, S> nextNodeToBeEvaluated;
```
globalworming, line 36: this was wrong on many levels. why are there Queues created when the next node is always stored in an instance variable?

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -28,80 +29,120 @@
...
-            bidirectionalActions = new BasicBidirectionalActions<>(fromInitialStatePart(), fromGoalStatePart());
-            return bidirectionalActions;
+        if (originalProblem.isGoalState(originalProblem.initialState())) {
+            return Collections.emptyList();
         }
-        Node<A, S> revNode = newRootNode(reverseProblem.initialState(), 0);
-        Queue<Node<A, S>> front = newFIFOQueue(node);
-        Queue<Node<A, S>> back = newFIFOQueue(revNode);
-        Map<S, Node<A, S>> exploredFront = newExploredMap(node.state());
-        Map<S, Node<A, S>> exploredBack = newExploredMap(revNode.state());
```
globalworming, line 52: there is no need for the explored states to be a map.

core/src/main/java/aima/core/search/basic/uninformed/BidirectionalSearch.java
```diff
@@ -28,80 +29,120 @@
...
 
-        while (!front.isEmpty() && !back.isEmpty()) {
```
globalworming, line 54: obsolete `while` as front is never filled with more than the initial node.

globalworming:
sorry if the two commits bother you. maybe merge #294 first, then this should be a single commit.. I guess

core/src/main/java/aima/core/search/api/SearchForActionsBidirectionallyFunction.java
```diff
@@ -1,12 +1,14 @@
...
 
 /**
  * @author manthan
  */
 @FunctionalInterface
-public interface SearchForActionsBidirectionallyFunction<A, S> extends BiFunction<Problem<A, S>, Problem<A, S>, BidirectionalActions<A>> {
-    @Override
-    BidirectionalActions<A> apply(Problem<A, S> originalProblem, Problem<A, S> reverseProblem);
+public interface SearchForActionsBidirectionallyFunction<A, S> extends Function<Pair<Problem<A,
+    S>, Problem<A, S>>, List<A>> {
```
ctjoreilly, line 13: I am not sure of the advantage of switching from extending BiFunction to Function. I realize it works in the test case scenario in that the factory returns a Pair and therefore its just simple to pass that to the function. However, in general and in this case I don't think it makes sense to squash the arguments into a Pair structure.

As regards your concerns about BidirectionalActions. I take your point that what you want, in most cases, is just the solution - i.e. the list of actions from the start to the goal. However, this assumes traversal always from 1 direction. This may not always be the case. For example, what if you want to develop an route finding application that gives directions to two people to meet in the middle. Or more simplify if you want to build a GUI demonstrating how the algorithm works. Therefore, for now anyway, I prefer that this API stay as is.

ctjoreilly (changes requested):
Hi @globalworming,

If you are good with reverting the code to work with the pre-existing SearchForActionsBidirectionallyFunction API (see comments) and can sort out the merge conflict I will be happy to merge in this request as it addresses the defect(s) you identified with the extended test cases you added.

core/src/main/java/aima/core/search/api/SearchForActionsBidirectionallyFunction.java
globalworming: > I am not sure of the advantage of switching from extending BiFunction to Function

there is no real benefit, I can revert that without problems.

> route finding application that gives directions to two people to meet in the middle

ok, if you find that useful, I won't argue. I'd still rename `BidirectionalActions` to something like `BiDirectionalSearchSolution` because the name is pretty misleading imho

globalworming:
points taken and will be fixed. most likely in another pull request.

core/src/main/java/aima/core/search/api/SearchForActionsBidirectionallyFunction.java
ctjoreilly: Lets go with 'BidirectionalSearchSolution' (i.e.no uppercase 'D').

Thanks

Ciaran

ctjoreilly:
Closing as addressed in pull request #308.

PR #299, opened 2017-03-31
Optimality of AStarSearch 

kanishk1225:
Part of the book
Optimality of A*
As we mentioned earlier, A* has the following properties: the tree-search version of A* is optimal if h(n) is admissible, while the graph-search version is optimal if h(n) is consistent. 
We show the second of these two claims since it is more useful. The argument essentially mirrors the argument for the optimality of uniform-cost search, with g replaced by f—just as in the A* algorithm itself.
The first step is to establish the following: if h(n) is consistent, then the values of f(n) along any path are nondecreasing. The proof follows directly from the definition of consistency. Suppose n' is a successor of n; then g(n')=g(n) + c(n, a, n') for some action a, and we have 
f(n') = g(n') + h(n') = g(n) + c(n, a, n') + h(n') ≥ g(n) + h(n) = f(n) .
The next step is to prove that whenever A* selects a node n for expansion, the optimal path to that node has been found. Were this not the case, there would have to be another frontier node n' on the optimal path from the start node to n, by the graph separation property of Figure 3.9; because f is nondecreasing along any path, n' would have lower f-cost than n and would have been selected first.
figure 3 9

### DIFF
--- a/core/src/main/java/aima/core/search/basic/informed/AStarSearch.java
+++ b/core/src/main/java/aima/core/search/basic/informed/AStarSearch.java
@@ -71,11 +71,13 @@ public class AStarSearch<A, S> implements SearchForActionsFunction<A, S> {
 				if (!(explored.contains(child.state()) || containsState(frontier, child.state()))) {
 					// frontier <- INSERT(child, frontier)
 					frontier.add(child);
-				} // else if child.STATE is in frontier with higher COST then
+				} 
+				//optimality of A*
+				/*// else if child.STATE is in frontier with higher COST then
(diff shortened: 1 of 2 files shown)


kanishk1225:
@ctjoreilly please review this.

kanishk1225:
Could you tell me how can I bring in optimality condition in A* algorithm

ctjoreilly:
Could you show me an example of where you think its not producing an
optimal answer.

kanishk1225:
sorry for bothering you. I had a deeper look into optimality condition and found that code has taken care of it.

Issue #245, opened 2017-01-12
GibbsAsk.ask() ignores evidence


andredd:
There might be an error in the GibbsAsk implementation. Evidence for child nodes is ignored. RejectionSampling and Likelyhoodweighting work as expected. Test follows:

        // create two nodes: parent and child with an arc from parent to child 
        RandomVariable rvParent = new RandVar("Parent", new BooleanDomain());
        RandomVariable rvChild = new RandVar("Child", new BooleanDomain());
        FullCPTNode nodeParent = new FullCPTNode(rvParent, new double[] { 0.7, 0.3 });
        new FullCPTNode(rvChild, new double[] { 0.8, 0.2, 0.2, 0.8 }, nodeParent);

        // create net
        BayesNet net = new BayesNet(nodeParent);

        // query parent probability
        RandomVariable[] rvX = new RandomVariable[] { rvParent };

        // ...given child evidence (true)
        AssignmentProposition[] propE = new AssignmentProposition[] { new AssignmentProposition(rvChild, Boolean.TRUE) };

        // sample with LikelihoodWeighting: result should be correct
        CategoricalDistribution samplesLW = new LikelihoodWeighting().ask(rvX, propE, net, 1000);
        assertEquals(0.9, samplesLW.getValue(Boolean.TRUE), 0.1);

        // sample with RejectionSampling: result should be correct
        CategoricalDistribution samplesRS = new RejectionSampling().ask(rvX, propE, net, 1000);
        assertEquals(0.9, samplesRS.getValue(Boolean.TRUE), 0.1);

        // sample with GibbsAsk: result is weird
        CategoricalDistribution samplesGibbs = new GibbsAsk().ask(rvX, propE, net, 1000);
        assertEquals(0.9, samplesGibbs.getValue(Boolean.TRUE), 0.1);

ngrj93:
After studying the gibbs sampling code and doing an extensive debugging run of the above submitted code I noted that the probability over markov blanket being calculated is over the same **event** regardless of the new values being assigned to Xi over the loop. I believe that is the cause of the problem although I may be wrong.

The normalized probability values being returned are [0.56, 0.24] in the case of {parent = true, child = true} for the above test case. I believe it should be [0.56, 0.06]. Could this be verified?

```
public static double[] mbDistribution(Node Xi,
			Map<RandomVariable, Object> event) {
	FiniteDomain fd = (FiniteDomain) Xi.getRandomVariable().getDomain();
[...]

PR #305, opened 2017-04-02
Fix Gibbs sampling bug and add supporting testcases (Issue #245)

ngrj93:
This is in reference to open issue #245

I performed a big debugging run and have identified and fixed the following:

**Updates**
1. **_Problem_**
Gibbs Sampling was producing the wrong results for the test case (Issue #245). 
**_Cause_**
The event under consideration when computing the distribution under the Markov blanket remained 
static throughout all the computations for the child nodes regardless of the change in the domain of 
the random variable of the parent.
**_Solution_**
Varied the domain value for the parent variable in the event to compute child Markov blanket probabilities.

2. **_Problem_** 
Current test case alone is not a valid check of the Gibbs sampling routine. The mock randomizer values severely skewed the resultant distribution. Since the routine was run for only 3 iterations, the results are inconclusive (as number of iterations increased, the probabilities tended to (0.5, 0.5) . Moreover the bug above was giving the wrong answer anyway.
**_Solution_** 
Changed test case iterations to 1000 runs to get better results. Manually computed expected posterior probabilities and verified the algorithm was returning the same.

### DIFF
--- a/aima-core/src/main/java/aima/core/probability/util/ProbUtil.java
+++ b/aima-core/src/main/java/aima/core/probability/util/ProbUtil.java
@@ -1,5 +1,6 @@
 package aima.core.probability.util;
 
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


ctjoreilly:
Good catch and nice fix.

Thanks

Ciaran

PR #307, opened 2017-04-02
bidirectional search

globalworming:
more unittests, all cases in this environment should work


globalworming:
sorry, conflicts...

PR #310, opened 2017-04-02
Optimize toString() methods with streams

ngrj93:
**Updates**

- Shortened toString() methods using Java 8 streams


ctjoreilly:
These are reasonable/clean updates. However, I would like to avoid these kinds of changes on the AIMA3e branch. Even though it now support JDK 8 the aim is to apply these kinds of changes/updates on the AIMA4e branch as we re-implement pre-existing algorithms. Therefore, I will not be merging this pull request. As an aside, using:

May also simplify the code further.

Thanks

Ciaran

Issue #52, opened 2015-03-12
Add support for general purpose environment simulator


GoogleCodeExporter:
```
Note: This capability is referred to in pages 45-46 of AIMA3e -

"The code repository associated with this book (aima.cs.berkeley.edu)
includes implementations of a number of environments, together with a
general-purpose environment simulator that places one or more agents in a
simulated environment, observes their behavior over time, and evaluates
them according to a given performance measure. Such experiments are
often carried out not for a single environment but for many environments
drawn from an environment class. For example, to evaluate a taxi driver in
simulated ENVIRONMENT traffic, we would want to CLASS run many simulations
with different traffic, lighting, and weather conditions. If we designed
the agent for a single scenario, we might be able to take advantage of
specific properties of the particular case but might not identify a good
design for driving in general. For this reason, the code repository also
includes an environment generator ENVIRONMENT for each environment
GENERATOR class that selects particular environments (with certain
likelihoods) in which to run the agent. For example, the vacuum environment
generator initializes the dirt pattern and agent location
randomly. We are then interested in the agent’s average performance over
the environment class. A rational agent for a given environment class
maximizes this average performance. Exercises 2.8 to 2.13 take you through
the process of developing an environment class and evaluating various
agents therein."
```

Original issue reported on code.google.com by `ctjoreilly@gmail.com` on 16 Feb 2010 at 6:46

ctjoreilly:
Ideally, this should be developed in the 'extra' module of the AIMA4e branch, i.e.:

aima.extra.environment

with sub-packages:
.simulator // for general purpose simulator
.generator

where an initial submission for this issue should focus on the Vacuum world and Map2D environments from the core.

jianxinliu:
Can I apply for solving this issue in GSoC 2017? @ctjoreilly

ctjoreilly:
Yes but it may be tricky at this stage of the AIMA4e's development as not all the algorithms are in place/stabilized.

jianxinliu:
Thank you for your reply.I have mentioned it in my proposal. @ctjoreilly

PR #311, opened 2017-04-02
Initial structure for general purpose environment #52

AdrianBZG:
Initial structure to add support for general purpose environment simulator (#52):

- EnvironmentMap class that represents a full map extension that is composed by MapSector, each MapSector has MapChunks that can contain 'n' elements. The Map generates by on-demand in-memory MapSector(s), this way we can have a simulation environment with unlimited space (only limited by the machine memory). Some memory optimization, e.g.: delete a MapSector when it hasn't been used for a while, are a future feature for the Map.

- The EnvironmentMap can be generated empty or using the Generator interface, I have implemented an example generator, that is a Voronoi Noise (or Cell Noise) and a NullGenerator.

- The general structure follows the MVC pattern; I have added some example views: a generic ObjectView and a BlockView (that represents an obstacle, wall...)

### DIFF
--- a/core/src/main/java/aima/core/util/datastructure/Tuple.java
+++ b/core/src/main/java/aima/core/util/datastructure/Tuple.java
@@ -1,0 +1,82 @@
+package aima.core.util.datastructure;
+
+/**
+ * Created by @AdrianBZG on 2/04/17.
+ */
+
+import java.util.function.Function;
+
+/**
+ * Util class: a parametric tuple type
+ */
+public class Tuple<A, B> {
+    private A fst;
+    private B snd;
+
+    public Tuple(A fst, B snd) {
+        this.fst = fst;
+        this.snd = snd;
+    }
+
+    public A getWidth() {
+        return fst;
+    }
+
+    public B getHeight() {
+        return snd;
+    }
+
+    public A getX() {
+        return fst;
+    }
+
+    public B getY() {
+        return snd;
+    }
+
+    public A getFst() {
+        return fst;
+    }
+
+    public void setFst(A fst) {
+        this.fst = fst;
+    }
+
+    public B getSnd() {
+        return snd;
+    }
+
+    public void setSnd(B snd) {
+        this.snd = snd;
+    }
+
+    public void chgFst(Function<A, A> fun) {
+        fst = fun.apply(getFst());
+    }
+
+    public void chgSnd(Function<B, B> fun) {
+        snd = fun.apply(getSnd());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        Tuple<?, ?> tuple = (Tuple<?, ?>) o;
+
+        return tuple.getX() == this.getX() && tuple.getY() == this.getY();
+    }
+
+    @Override
+    public int hashCode() {
+        int result = fst != null ? fst.hashCode() : 0;
+        result = 31 * result + (snd != null ? snd.hashCode() : 0);
+        return result;
+    }
+
+    @Override
+    public String toString () {
+        return getX() + ", " + getY();
+    }
+}
--- a/extra/src/main/java/aima/extra/environment/model/map/EnvironmentMap.java
+++ b/extra/src/main/java/aima/extra/environment/model/map/EnvironmentMap.java
@@ -1,0 +1,199 @@
+package aima.extra.environment.model.map;
+
+/**
+ * Created by @AdrianBZG on 2/04/17.
+ */
+
+import aima.core.util.datastructure.Tuple;
+import aima.extra.environment.model.map.generator.IGenerator;
+import aima.extra.environment.model.map.object.Block;
+import aima.extra.environment.model.map.object.MapObject;
+
+import java.util.HashMap;
+import java.util.Optional;
+
+/**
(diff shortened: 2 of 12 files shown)


core/src/main/java/aima/core/util/datastructure/Tuple.java
```diff
@@ -0,0 +1,82 @@
...
+ * Created by @AdrianBZG on 2/04/17.
+ */
+
+import java.util.function.Function;
+
+/**
+ * Util class: a parametric tuple type
+ */
+public class Tuple<A, B> {
+    private A fst;
```
globalworming, line 13: I'd write `first` instead of `fst`. The abbreviation isn't helping anyone.
see: clean code http://www.itiseezee.com/?p=83

core/src/main/java/aima/core/util/datastructure/Tuple.java
```diff
@@ -0,0 +1,82 @@
...
+    public Tuple(A fst, B snd) {
+        this.fst = fst;
+        this.snd = snd;
+    }
+
+    public A getWidth() {
+        return fst;
+    }
+
+    public B getHeight() {
```
globalworming, line 25: A Tuple does not have a height property. If you need a height, create a `class Dimensions extends Tuple<Width, Height>`

extra/src/main/java/aima/extra/environment/model/map/EnvironmentMap.java
```diff
@@ -0,0 +1,199 @@
...
+public class EnvironmentMap implements Cloneable {
+
+    private static int CHUNK_SIZE = 32;
+
+    /**
+     * Partial Map with zones explored
+     */
+    HashMap<MapSector, MapChunk> map = new HashMap<>();
+
+    Optional<Tuple<Integer, Integer>> dimensions = Optional.empty();
```
globalworming, line 33: using `Optional` as properties is considered bad practice, could you do anything else here? Like using `null`?
https://stackoverflow.com/a/29033935/1359903
https://www.insaneprogramming.be/article/2017/01/06/the-usages-of-optional/

extra/src/main/java/aima/extra/environment/model/map/MapChunk.java
```diff
@@ -0,0 +1,74 @@
...
+public class MapChunk {
+    /**
+     * Represent chunk size
+     */
+    private int size;
+
+    /**
+     * Represent information about contents in the chunk
+     */
+    HashMap<Integer, MapObject> chunk = new HashMap<>();
```
globalworming, line 20: Use the `Map` interface, or is there any real need for this to be a `HashMap`?

extra/src/main/java/aima/extra/environment/model/map/object/MapObject.java
```diff
@@ -0,0 +1,82 @@
...
+     * Internal use to have a list of all object available
+     */
+    private static ArrayList<MapObject> mapObjects = new ArrayList<>();
+
+    private Tuple<Integer, Integer> objectPosition = new Tuple<>(0, 0);
+
+
+    static {
+        // No objects at the moment so the Array is empty
+        mapObjects = new ArrayList<>(Arrays.asList());
```
globalworming, line 25: wtf? `mapObjects` gets an empty list in line 18, why here again? and why `Arrays.asList()` that is redundant too.

extra/src/main/java/aima/extra/environment/model/map/object/MapObject.java
```diff
@@ -0,0 +1,82 @@
...
+ * Created by @AdrianBZG on 2/04/17.
+ */
+
+import aima.core.util.datastructure.Tuple;
+import aima.extra.environment.view.object.ObjectView;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public abstract class MapObject implements Cloneable {
```
globalworming, line 13: I'd go with a component based approach, but that's just an opinion. http://gameprogrammingpatterns.com/component.html
this would remove mostly anything from the `MapObject` class and instead place some logic in the interfaces and create a `class Obstacle extends MapObject inplements Position, Render, ReadOnlyOptions ...`.

extra/src/main/java/aima/extra/environment/model/map/generator/VoronoiNoise.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.util.Random;
+
+/**
+ * Created by @AdrianBZG on 2/04/17.
+ *
+ * Just an example noise to generate a 2D Map environment: Voronoi Noise (also
+ * known as Cell Noise)
+ */
+public class VoronoiNoise implements IGenerator {
```
globalworming, line 11: I find the lack of unittests disturbing. Could you improve on that?

globalworming (changes requested):
Some clean code issues, one general architecture issue and no unit tests.

core/src/main/java/aima/core/util/datastructure/Tuple.java
globalworming: or if you want to avoid `Tuple` and use some Java Standard: `class Dimensions extends AbstractMap.SimpleEntry<Width, Height>`

norvig:
There are some nice utilities here, but also some replication of standard Java functionality, and there are no unit tests. So I'm going to close this PR.

Issue #192, opened 2016-03-22
AIMA4e reimplement Fig 18.5 Decision-Tree-Learner from AIMA3e branch


ctjoreilly:

Kmayankkr:
I would like to write this code, taking reference from the already written AIMA4e algorithms to get an idea of the writing style. If you would like, please point out any major difference between 4e and 3e. @ctjoreilly

shantanusinghal:
@Kmayankkr Are you still working on this implementation?

shantanusinghal:
I've created pull request #306. Please review.

PR #313, opened 2017-04-03
Added Decision-Tree-Learner from Fig 18.5

shantanusinghal:
Issue #192 | Added decision tree learner. While the API supports continuous valued attributes, I've only included the implementation for nominal attributes.

I've refactored the code to make it more extensible and enhanced the test-suite to extend coverage and make it more readable.

Looking forward to some feedback on my work, especially on the test-cases since I couldn't form a cohesive understanding of the testing-styles used by the community.


shantanusinghal:
I've added an Attribute implementation that supports numeric data types and used the approach outlined on page 707 to split the numeric attribute once.

Please let me know if the DataSet API needs to be enhanced to read from files and such. I've only designed a simple dataset that suffices for unit testing.

Also, I didn't squashed these two commits because they correspond to different issue tickets, let me know if they still need to be.

PR #304, opened 2017-04-02
Add support for complex actions

Manthan-R-Sheth:
The method of reversing and adding to the list is changed to support complex actions in the problem.
1. ```fromMeetingStateToInitialStateList``` and ```fromMeetingStateToGoalStateList``` are made to achieve the purpose.
1. InitialState and GoalState are included in the fromGoalStateToInitialState and fromInitialStateToGoalState respectively.

PR #314, opened 2017-04-03
Add support for complex actions and handle tests

Manthan-R-Sheth:
With this PR, the implementation of ```BidirectionalSearch``` is handled (in the form of BidirectionalSearchMRS) 
Support for complex actions is added, though not tested.


Manthan-R-Sheth:
This is with regards to #304, I have made the necessary changes. Some cases for node adjacency are handled to get the desired test cases passed.
This support can be tested with complex actions, which I plan to do next, if this is fine to get through.

Issue #34, opened 2015-03-12
Add implementation of - Fig 10.9  Graphplan


GoogleCodeExporter:
```
Fig 10.9  Graphplan


```

Original issue reported on code.google.com by `ctjoreilly@gmail.com` on 23 Oct 2009 at 2:46

smsubham:
Is this issue still open. If yes would like to work on it.

ctjoreilly:
Yes, still open.

smsubham:
Sir i have submitted a pull request for this issue , please review it at earliest and provide your valuable feedback.
https://github.com/aimacode/aima-java/pull/212

ajayvers:
@ctjoreilly  #212 is closed, does it mean can we close this issue, or still work is there ?

ctjoreilly:
It was closed as it wasn't a complete submission. There is currently no GraphPlan implementation on the AIMA3e branch, see index of implemented algorithms on main page.

letyrodridc:
I'm working on this issue. I hope to send a pull request soon.

samagra14:
This issue was resolved in #407. Therefore closing this issue.

PR #317, opened 2017-04-17
Added Graphplan implementation (Figure 10.9)

mgrenander:
Issue [#34](https://github.com/aimacode/aima-java/issues/34): Added implementation of the GraphPlan algorithm. I modeled most of my documentation syntax off of the SATPlan implementation. There is also a test case from the textbook, the spare tire example described on page 370 and 383 - 385 in AIMA3e.

If there's any feedback please let me know.


Manthan-R-Sheth:
@mgrenander, if you could commit only the files you changed/created instead of the entire set of files. It will be easier to review your changes then.
Thanks.

mgrenander:
Hi, 
Sorry about that - I'm not quite sure what was the cause of all the files being included in the commit. I've modified my fork to only include my files and I also put the implementation into a single commit.

PR #319, opened 2017-06-01
Initial implementation of ProbabilityNumber

ngrj93:
**Added**

- ProbabilityNumber interface 
- Implementations of ProbabilityNumber
  - DoubleType (standard double implementation)
  - BigNumberType (BigDecimal implementation)
  - LogType (Logspace computations)
  - RationalType (Rational space implementation)

- Tests added
  - DoubleTest (for DoubleType)
  - BigNumberTest (for BigNumberType)

### DIFF
--- a/extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
+++ b/extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
@@ -1,0 +1,381 @@
+package aima.extra.probability;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.math.MathContext;
+import java.math.RoundingMode;
+
+/**
+ * Implementation based on the underlying BigDecimal datatype. Supports higher
+ * precision than DoubleProbabilityNumber (DECIMAL128 precision by default, but
+ * can go upto UNLIMITED). The class is immutable.
+ * 
+ * @author Nagaraj Poti
+ *
+ */
+public class BigDecimalProbabilityNumber implements ProbabilityNumber, Comparable<ProbabilityNumber> {
+
+	// Constants
+
+	/**
+	 * IEEE 754R Decimal128 format - Precision of 34 digits and a rounding mode
+	 * of HALF_EVEN. HALF_EVEN rounding mode statistically minimizes cumulative
+	 * error when applied repeatedly over a sequence of calculations.
+	 */
+	private static final Integer DEFAULT_PRECISION = MathContext.DECIMAL128.getPrecision();
+
+	private static final RoundingMode DEFAULT_PRECISION_ROUNDING_MODE = MathContext.DECIMAL128.getRoundingMode();
+
+	/**
+	 * Maximum integer value that can be accomodated as exponent in the
+	 * BigDecimal pow() function
+	 */
+	private static final int EXPONENT_MAX = 999999999;
+
+	// Internal fields
+
+	private BigDecimal value;
+
+	/**
+	 * MathContext of value.
+	 */
+	private MathContext precisionMathContext = new MathContext(DEFAULT_PRECISION, DEFAULT_PRECISION_ROUNDING_MODE);
+
+	private BigDecimal BIG_DECIMAL_ZERO = new BigDecimal(0, precisionMathContext);
+
+	private BigDecimal BIG_DECIMAL_ONE = new BigDecimal(1, precisionMathContext);
+
+	// Constructors
+
+	/**
+	 * Construct a BigDecimalProbabilityNumber from double type.
+	 * BigDecimal.valueOf() builds from a double type based on its canonical
+	 * representation. Precision of double type is at max 15.95 (~15).
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 */
+	public BigDecimalProbabilityNumber(double value) {
+		if (value < 0 || value > 1) {
+			throw new IllegalArgumentException();
+		}
+		this.value = new BigDecimal(Double.toString(value), precisionMathContext);
+	}
+
+	/**
+	 * Construct a BigDecimalProbabilityNumber from BigDecimal type.
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 */
+	public BigDecimalProbabilityNumber(BigDecimal value) {
+		if (null == value || value.compareTo(BIG_DECIMAL_ZERO) == -1 || value.compareTo(BIG_DECIMAL_ONE) == 1) {
+			throw new IllegalArgumentException();
+		}
+		this.value = new BigDecimal(value.toString(), precisionMathContext);
+	}
+
+	/**
+	 * Construct a BigDecimalProbabilityNumber from BigDecimal type.
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 * 
+	 * @param mc
+	 *            MathContext to be associated with value.
+	 */
+	public BigDecimalProbabilityNumber(BigDecimal value, MathContext mc) {
+		precisionMathContext = mc;
+		setConstants();
+		if (value.compareTo(BIG_DECIMAL_ZERO) == -1 || value.compareTo(BIG_DECIMAL_ONE) == 1) {
+			throw new IllegalArgumentException();
+		}
+		this.value = new BigDecimal(value.toString(), precisionMathContext);
+	}
+
+	// Private methods
+
+	/**
+	 * Initialize constants BIG_DECIMAL_ZERO and BIG_DECIMAL_ONE according to
+	 * precisionMathContext that is defined.
+	 */
+	private void setConstants() {
+		BIG_DECIMAL_ZERO = new BigDecimal(0, precisionMathContext);
+		BIG_DECIMAL_ONE = new BigDecimal(1, precisionMathContext);
+	}
+
+	/**
+	 * Covert other implementations of the ProbabilityNumber interface to a
+	 * BigDecimalProbabilityNumber.
+	 * 
+	 * @param that
+	 *            BigDecimalProbabilityNumber.
+	 */
+	private BigDecimalProbabilityNumber toInternalType(ProbabilityNumber that) {
+		if (that instanceof BigDecimalProbabilityNumber) {
+			return (BigDecimalProbabilityNumber) that;
+		}
+		return new BigDecimalProbabilityNumber(that.getValue(), that.getMathContext());
+	}
+
+	/**
+	 * Compare two BigDecimal values upto minimum precision.
+	 * 
+	 * @param first
+	 *            of BigDecimal type.
+	 * @param second
+	 *            of BigDecimal type.
+	 * 
+	 * @return 1 if first > second, 0 if first == second, -1 if first < second.
+	 */
+	private int compareBigDecimal(BigDecimal first, MathContext mc1, BigDecimal second, MathContext mc2) {
+		if (mc1.getPrecision() == mc2.getPrecision()) {
+			return first.compareTo(second);
+		}
+		int minPrecision = getMinPrecision(mc1.getPrecision(), mc2.getPrecision());
+		BigDecimal valueA = new BigDecimal(first.toString(), new MathContext(minPrecision, mc1.getRoundingMode()));
+		BigDecimal valueB = new BigDecimal(second.toString(), new MathContext(minPrecision, mc2.getRoundingMode()));
+		return valueA.compareTo(valueB);
+	}
+
+	/**
+	 * Return the minimum of two precision values. If any of the precision
+	 * values is 0, the maximum value of the two is returned.
+	 * 
+	 * @param precisionA
+	 * @param precisionB
+	 * @return
+	 */
+	private int getMinPrecision(int precisionA, int precisionB) {
+		if (precisionA == 0 || precisionB == 0) {
+			return Math.max(precisionA, precisionB);
+		} else {
+			return Math.min(precisionA, precisionB);
+		}
+	}
+
+	// Public methods
+
+	/**
+	 * @return value of BigDecimal type.
+	 */
+	@Override
+	public BigDecimal getValue() {
+		return value;
+	}
+
+	/**
+	 * @return precisionMathContext used by value set to DECIMAL128 precision
+	 *         and HALF_EVEN RoundingMode by default.
+	 */
+	@Override
+	public MathContext getMathContext() {
+		return precisionMathContext;
+	}
+
+	/**
+	 * Checks if the value is zero or not. The first check is an absolute check
+	 * upto MAX_PRECISION digits. If the check fails, an approximation check
+	 * based on the specified DEFAULT_SCALE and DEFAULT_SCALE_ROUNDING_MODE is
+	 * made.
+	 * 
+	 * @return true if zero, false otherwise.
+	 */
+	@Override
+	public boolean isZero() {
+		return (compareBigDecimal(this.value, this.precisionMathContext, BIG_DECIMAL_ZERO, this.precisionMathContext) == 0);
+	}
+
+	/**
+	 * Checks if the value is one. The first check is an absolute check upto
+	 * MAX_PRECISION digits. If the check fails, an approximation check based on
+	 * the specified DEFAULT_SCALE and DEFAULT_SCALE_ROUNDING_MODE is made.
+	 * 
+	 * @return true if one, false otherwise.
+	 */
+	@Override
+	public boolean isOne() {
+		return (compareBigDecimal(this.value, this.precisionMathContext, BIG_DECIMAL_ONE, this.precisionMathContext) == 0);
+	}
+
+	/**
+	 * Checks if argument implementing ProbabilityNumber interface is equal to
+	 * the value of the current BigDecimalProbabilityNumber. The check between
+	 * two ProbabilityNumbers involves a check only upto the minimum precision
+	 * of the two values, i.e min(this.precision(), that.precision()).
+	 * 
+	 * @param that
+	 *            the ProbabilityNumber type that is to be compared with this
+	 *            BigDecimalProbabilityNumber.
+	 *
+	 * @return true if this == that upto precision value specified by
+	 *         min(this.precision(), that.precision()), false otherwise.
+	 */
+	@Override
+	public boolean equals(ProbabilityNumber that) {
+		BigDecimalProbabilityNumber specificType = toInternalType(that);
+		return (compareBigDecimal(this.value, this.precisionMathContext, specificType.getValue(), specificType.getMathContext()) == 0);
+	}
+	
+	/**
+	 * Compare this with another ProbabilityNumber value (that).
+	 * 
+	 * @param that
+	 *            of type ProbabilityNumber.
+	 * 
+	 * @return 1 if this > that, 0 if this == that, -1 if this < that.
+	 */
+	@Override
+	public int compareTo(ProbabilityNumber that) {
+		BigDecimalProbabilityNumber second = toInternalType(that);
+		return compareBigDecimal(this.value, this.precisionMathContext, second.getValue(), second.getMathContext());
+	}
+
+	/**
+	 * Add a BigDecimalProbabilityNumber to this BigDecimalProbabilityNumber and
+	 * return a new BigDecimalProbabilityNumber.
+	 * 
+	 * @param that
+	 *            the BigDecimalProbabilityNumber that is to be added to this
+	 *            BigDecimalProbabilityNumber.
+	 * 
+	 * @return a new BigDecimalProbabilityNumber that is the result of addition.
+	 */
+	@Override
+	public ProbabilityNumber add(ProbabilityNumber that) {
+		BigDecimalProbabilityNumber addend = toInternalType(that);
+		int minPrecision = getMinPrecision(this.precisionMathContext.getPrecision(), addend.getMathContext().getPrecision());
+		MathContext resultMathContext = new MathContext(minPrecision, DEFAULT_PRECISION_ROUNDING_MODE);
+		return new BigDecimalProbabilityNumber(value.add(addend.getValue(), resultMathContext), resultMathContext);
+	}
+
+	/**
+	 * Subtract a BigDecimalProbabilityNumber from this
+	 * BigDecimalProbabilityNumber and return a new BigDecimalProbabilityNumber.
+	 * 
+	 * @param that
+	 *            the BigDecimalProbabilityNumber that is to be subtracted from
+	 *            this BigDecimalProbabilityNumber.
+	 * 
+	 * @return a new BigDecimalProbabilityNumber that is the result of
+	 *         subtraction.
+	 */
+	@Override
+	public ProbabilityNumber subtract(ProbabilityNumber that) {
+		BigDecimalProbabilityNumber subtrahend = toInternalType(that);
+		int minPrecision = getMinPrecision(this.precisionMathContext.getPrecision(),
+				subtrahend.getMathContext().getPrecision());
+		MathContext resultMathContext = new MathContext(minPrecision, DEFAULT_PRECISION_ROUNDING_MODE);
+		return new BigDecimalProbabilityNumber(value.subtract(subtrahend.getValue(), resultMathContext),
+				resultMathContext);
+	}
+
+	/**
+	 * Multiply a BigDecimalProbabilityNumber with this
+	 * BigDecimalProbabilityNumber and return a new BigDecimalProbabilityNumber.
+	 * 
+	 * @param that
+	 *            the BigDecimalProbabilityNumber that is to be multiplied to
+	 *            this BigDecimalProbabilityNumber.
+	 * 
+	 * @result a new BigDecimalProbabilityNumber that is the result of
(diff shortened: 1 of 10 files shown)


extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -0,0 +1,40 @@
...
+import java.math.BigInteger;
+
+/**
+ * This interface is implemented by the various ProbabilityNumber
+ * implementations.
+ * 
+ * @author Nagaraj Poti
+ *
+ */
+public interface ProbabilityNumber {
```
ctjoreilly: If you look at the Wiki: https://en.wikipedia.org/wiki/Probability
a 'Probability is quantified as a number between 0 and 1' and I think this is what the interface should represent externally (how implementations internally represent that should be hidden, e.g. logspace). 

This would remove the need for the following methods:
* int signum()
* boolean isPosititive()
* boolean isNegative()
* ProbabilityNumber abs()
* ProbabilityNumber negate()
* ProbabilityNumber invert()

I would add the following methods:
* isOne()
* MathContext getMathContext() - so people can get an idea of the precision information used internally.
* BigDecimal getValue(); that will always return a value between 0 and 1inclusive, which is a standardized representation of the internal implementation (e.g. the logspace implemenation would output the exponentiated value not its internal log value).
* boolean sumsToOne(Iterable<ProbabilityNumber> allProbabilities) - this relates to aima3es need to handle rounding thresholds, see: https://github.com/aimacode/aima-java/blob/AIMA3e/aima-core/src/main/java/aima/core/probability/ProbabilityModel.java.

Minor cleanup:
* On an interface there is no need to declare methods 'public' as this is implicit due to being an interface.

As regards implementations of this interface. For now have them only work with their own types (i.e. don't try to mix Log representations with Double representations for now). i.e. have a method:

private ImplementingType toInternalType(ProbabilityNumber pn) {
}

which in a method you would use like this:

@Override
public ProbabilityNumber add(ProbabilityNumber pnAddend) {
    ImplementingType added = totoInternalType(pnAddend);
    ...
}

Later we can decide if we want to be able to mix different implementation types and this would only require updating the 'toInternalType' method with a common shared.

ctjoreilly (commented):
Hi,

This is a good start. I am going to focus this initial review exclusively on the ProbabilityNumber interface that you have proposed, as this is the basis for everything else that will be developed. See comment associated with that file for details.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
ngrj93: I have made some significant changes as per the updated requirements. I have added tests for all the four implementations as well.

ngrj93:
Updates 

- Removed booleanApproximation parameter from functions and made toInternalType private.

- First check is always for absolute precision, otherwise approximate precision is utilised.

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Implementation based on the underlying BigDecimal datatype. Supports higher
+ * precision than DoubleType (DECIMAL128 precision by default, but can go upto
+ * UNLIMITED). The class is immutable.
+ * 
+ * @author Nagaraj Poti
+ *
+ */
+public class BigNumberType implements ProbabilityNumber {
```
ctjoreilly: Recommend renaming to BigDecimalProbabilityNumber.

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	/**
+	 * Scale will be used for comparing two BigDecimal elements for equality
+	 * (approximately 1e-8 DEFAULT_ROUNDING_THRESHOLD)
+	 */
+	private static final Integer DEFAULT_SCALE = 7;
+
+	/**
+	 * Choose consistent rounding scheme when checking scaled values.
+	 */
+	private static final RoundingMode DEFAULT_SCALE_ROUNDING_MODE = RoundingMode.HALF_UP;
```
ctjoreilly: I am not sure the DEFAULT_SCALE and DEFAULT_SCALE_ROUNDING_MODE are required for comparing. Essentially, these values are implied by the lower precision of the two values being compared (i.e. the internal BigDecimal's MathContext).

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+
+	/**
+	 * Maximum integer value that can be accomodated as exponent in the
+	 * Math.pow() function
+	 */
+	private static final int EXPONENT_MAX = Integer.MAX_VALUE;
```
ctjoreilly: Careful, look at the implementation of BigDecimal.pow():
 if (n < 0 || n > 999999999)
    throw new ArithmeticException("Invalid operation");

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+
+	private static final BigDecimal BIG_DECIMAL_ZERO = new BigDecimal(0);
+
+	private static final BigDecimal BIG_DECIMAL_ONE = new BigDecimal(1);
+
```
ctjoreilly: Should the MathContext UNLIMITED note be used for these two constants ^^^

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	// Internal fields
+
+	private BigDecimal value;
+
+	private static MathContext approximationMathContext = new MathContext(DEFAULT_MAX_PRECISION,
+			DEFAULT_PRECISION_ROUNDING_MODE);
+
```
ctjoreilly: Again, not sure this is required ^

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -0,0 +1,40 @@
...
+	ProbabilityNumber multiply(ProbabilityNumber multiplicand);
+
+	ProbabilityNumber divide(ProbabilityNumber divisor);
+
+	ProbabilityNumber pow(BigInteger exponent);
+
+	ProbabilityNumber pow(int exponent);
+
+	boolean sumsToOne(Iterable<ProbabilityNumber> allProbabilities);
+
```
ctjoreilly: Looks good. Recommend adding JavaDoc and also a discussion on the use of getMathContext() is the class javadoc, in particular as it pertains to UNLIMITED precision and the drawbacks of using.

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	/**
+	 * Construct a BigNumberType from double type. BigDecimal.valueOf() builds
+	 * from a double type based on its canonical representation.
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 */
+	public BigNumberType(double value) {
+		this.value = BigDecimal.valueOf(value);
+	}
```
ctjoreilly: ^ Check the value passed is a legal argument (i.e. in the interval [0,1]

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	/**
+	 * Construct a BigNumberType from a primitive long type.
+	 * 
+	 * @param value
+	 *            to be assigned to BigNumberType value.
+	 */
+	public BigNumberType(long value) {
+		this.value = new BigDecimal(value);
+	}
+
```
ctjoreilly: ^ not required.

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	/**
+	 * Construct a BigNumbertype from a string type.
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 */
+	public BigNumberType(String value) {
+		this.value = new BigDecimal(value);
+	}
+
```
ctjoreilly: ^ Same argument assertion checks apply to these constructors as well with respect to [0,1].

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	/**
+	 * @return approximationMathContext set to DECIMAL128 precision and
+	 *         HALF_EVEN RoundingMode by default.
+	 */
+	@Override
+	public MathContext getMathContext() {
+		return approximationMathContext;
+	}
+
```
ctjoreilly: ^ not correct, it should be the math context used by value.

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	@Override
+	public boolean isZero() {
+		if (value.compareTo(BIG_DECIMAL_ZERO) == 0) {
+			return true;
+		}
+		BigDecimal first = value.setScale(DEFAULT_SCALE, DEFAULT_SCALE_ROUNDING_MODE);
+		BigDecimal second = BIG_DECIMAL_ZERO.setScale(DEFAULT_SCALE, DEFAULT_SCALE_ROUNDING_MODE);
+		return (first.compareTo(second) == 0);
+	}
+
```
ctjoreilly: This is not how the comparisons should work. You want to use the mathcontext with the lowest precision and do the comparison that way. For efficiency check that if the precision on both is the same already you don't need to pick a lowest value and convert one of them from high to low (should all be supported by a private utility routine).

NOTE: this comment applies to the comparisons performed in the following methods as well.

extra/src/main/java/aima/extra/probability/BigNumberType.java
```diff
@@ -0,0 +1,340 @@
...
+	 * 
+	 * @return a new BigNumberType that is the result of division.
+	 */
+	@Override
+	public ProbabilityNumber divide(ProbabilityNumber that) {
+		BigNumberType divisor = toInternalType(that);
+		// Consider approximate value of 0 as well?
+		if (divisor.isZero()) {
+			// Throw exception (can't divide by 0)
+			return null;
```
ctjoreilly: throw an IllegalArgumentException here.

ctjoreilly (commented):
Initial set of review comments focused on ProbabilityNumber and BigNumberType.

ngrj93:
I have updated the implementations. Please have a look.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -0,0 +1,149 @@
...
+import java.math.BigInteger;
+import java.math.MathContext;
+
+/**
+ * ProbabilityNumber interface defines arithmetic specifically for probability
+ * values. A probability value is quantified between 0 and 1. Operations that
+ * violate the property of a probability value are not permitted. This interface
+ * is implemented by the various ProbabilityNumber implementations.
+ * 
+ * @author Nagaraj Poti
```
ctjoreilly, line 13: NOTE: That the interface is intended to be immutable.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -0,0 +1,149 @@
...
+	 * @return value of BigDecimal type.
+	 */
+	BigDecimal getValue();
+
+	/**
+	 * Get the MathContext object associated with the ProbabilityNumber object.
+	 * MathContext can be set for BigDecimalProbabilityNumber and
+	 * RationalProbabilityNumber via the constructor mechanism. The default
+	 * MathContext object associated with the implementations is of DECIMAL128
+	 * precision. It is possible to go upto an even higher precision all the way
```
ctjoreilly, line 31: the -> these. Not sure the defaults for these implementations need to be mentioned here but ok.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Check if two ProbabilityNumber values are equal.
+	 * 
+	 * @param that
+	 *            is a ProbabilityNumber.
+	 * 
+	 * @return true if equal, false otherwise.
+	 */
+	boolean equals(ProbabilityNumber that);
+
```
ctjoreilly, line 85: Not sure this is required on the interface (i.e.equals(ProbabilityNumber). Adding a class comment recommending that implementations implement overridden versions of equals() and hashCode() should be sufficient.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Add two ProbabilityNumber types and return a ProbabilityNumber.
+	 *
```
ctjoreilly, line 88: return a new ProbabilityNumber representing the result of the addition (update related methods in a similar way).

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+/**
+ * Implementation based on the underlying BigDecimal datatype. Supports higher
+ * precision than DoubleProbabilityNumber (DECIMAL128 precision by default, but
+ * can go upto UNLIMITED). The class is immutable.
+ * 
+ * @author Nagaraj Poti
+ *
+ */
+public class BigDecimalProbabilityNumber implements ProbabilityNumber, Comparable<ProbabilityNumber> {
+
```
ctjoreilly, line 17: The ProbabilityNumber interface already extends Comparable<ProbabilityNumber> so you shouldn't have to add to this implements lists as well (applies to other implementations as well).

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	/**
+	 * MathContext of value.
+	 */
+	private MathContext precisionMathContext = new MathContext(DEFAULT_PRECISION, DEFAULT_PRECISION_ROUNDING_MODE);
+
+	private BigDecimal BIG_DECIMAL_ZERO = new BigDecimal(0, precisionMathContext);
+
+	private BigDecimal BIG_DECIMAL_ONE = new BigDecimal(1, precisionMathContext);
+
```
ctjoreilly, line 47: It likely makes more sense to declare these two constants as static and to give them unlimited precision (this will save a lot of space and address inconsistencies across different instances with different precisions).

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	 * Construct a BigDecimalProbabilityNumber from BigDecimal type.
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 */
+	public BigDecimalProbabilityNumber(BigDecimal value) {
+		if (null == value || value.compareTo(BIG_DECIMAL_ZERO) == -1 || value.compareTo(BIG_DECIMAL_ONE) == 1) {
+			throw new IllegalArgumentException();
+		}
+		this.value = new BigDecimal(value.toString(), precisionMathContext);
```
ctjoreilly, line 75: No, you should use the precision of the BigDecimal given to you and not revert it to the default precision defined in the class.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+
+	// Internal fields
+
+	private BigDecimal value;
+
+	/**
+	 * MathContext of value.
+	 */
+	private MathContext precisionMathContext = new MathContext(DEFAULT_PRECISION, DEFAULT_PRECISION_ROUNDING_MODE);
+
```
ctjoreilly, line 43: precisionMathContext should not be required as this information is available on the `BigDecimal value` attribute.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	/**
+	 * Construct a BigDecimalProbabilityNumber from BigDecimal type.
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 */
+	public BigDecimalProbabilityNumber(BigDecimal value) {
+		if (null == value || value.compareTo(BIG_DECIMAL_ZERO) == -1 || value.compareTo(BIG_DECIMAL_ONE) == 1) {
+			throw new IllegalArgumentException();
```
ctjoreilly, line 73: Add a message to the IllegalArgumentException explaining why it is.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	 * 
+	 * @param mc
+	 *            MathContext to be associated with value.
+	 */
+	public BigDecimalProbabilityNumber(BigDecimal value, MathContext mc) {
+		precisionMathContext = mc;
+		setConstants();
+		if (value.compareTo(BIG_DECIMAL_ZERO) == -1 || value.compareTo(BIG_DECIMAL_ONE) == 1) {
+			throw new IllegalArgumentException();
+		}
```
ctjoreilly, line 92: ^ this is a common assertion check that could be moved into a private routine to help reduce duplicated code.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	// Private methods
+
+	/**
+	 * Initialize constants BIG_DECIMAL_ZERO and BIG_DECIMAL_ONE according to
+	 * precisionMathContext that is defined.
+	 */
+	private void setConstants() {
+		BIG_DECIMAL_ZERO = new BigDecimal(0, precisionMathContext);
```
ctjoreilly, line 103: This method should become unnecessary if you use the value's mathContext instead of a separate attribute.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+
+	/**
+	 * Covert other implementations of the ProbabilityNumber interface to a
+	 * BigDecimalProbabilityNumber.
+	 * 
+	 * @param that
+	 *            BigDecimalProbabilityNumber.
+	 */
+	private BigDecimalProbabilityNumber toInternalType(ProbabilityNumber that) {
+		if (that instanceof BigDecimalProbabilityNumber) {
```
ctjoreilly, line 115: Just a style comment here, your call if you want to adapt. I usually find it helpful to define all public methods first and then private ones at the end of the file. For methods that oreverride/implement an interface order them the same way as they are presented in the interface.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	 * @return 1 if first > second, 0 if first == second, -1 if first < second.
+	 */
+	private int compareBigDecimal(BigDecimal first, MathContext mc1, BigDecimal second, MathContext mc2) {
+		if (mc1.getPrecision() == mc2.getPrecision()) {
+			return first.compareTo(second);
+		}
+		int minPrecision = getMinPrecision(mc1.getPrecision(), mc2.getPrecision());
+		BigDecimal valueA = new BigDecimal(first.toString(), new MathContext(minPrecision, mc1.getRoundingMode()));
+		BigDecimal valueB = new BigDecimal(second.toString(), new MathContext(minPrecision, mc2.getRoundingMode()));
+		return valueA.compareTo(valueB);
```
ctjoreilly, line 138: Isn't the above unnecessarily creating two intermediate instances (expensively as well using toString) when at most you should only every require 1 and in some cases none if they start with the same precision.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	 *            BigDecimalProbabilityNumber.
+	 *
+	 * @return true if this == that upto precision value specified by
+	 *         min(this.precision(), that.precision()), false otherwise.
+	 */
+	@Override
+	public boolean equals(ProbabilityNumber that) {
+		BigDecimalProbabilityNumber specificType = toInternalType(that);
+		return (compareBigDecimal(this.value, this.precisionMathContext, specificType.getValue(), specificType.getMathContext()) == 0);
+	}
```
ctjoreilly, line 218: equals could delegate to compareTo and test for result == 0.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	 *            BigDecimalProbabilityNumber.
+	 * 
+	 * @return a new BigDecimalProbabilityNumber that is the result of addition.
+	 */
+	@Override
+	public ProbabilityNumber add(ProbabilityNumber that) {
+		BigDecimalProbabilityNumber addend = toInternalType(that);
+		int minPrecision = getMinPrecision(this.precisionMathContext.getPrecision(), addend.getMathContext().getPrecision());
+		MathContext resultMathContext = new MathContext(minPrecision, DEFAULT_PRECISION_ROUNDING_MODE);
+		return new BigDecimalProbabilityNumber(value.add(addend.getValue(), resultMathContext), resultMathContext);
```
ctjoreilly, line 249: Not sure DEFAULT_PRECISION_ROUNDING_MODE is the correct value to use here. I would see if the two values share the same rounding mode and if so use it. If different then use the one from this class as opposed to the argument value.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	 * 
+	 * @return a new BigDecimalProbabilityNumber that is the result of
+	 *         subtraction.
+	 */
+	@Override
+	public ProbabilityNumber subtract(ProbabilityNumber that) {
+		BigDecimalProbabilityNumber subtrahend = toInternalType(that);
+		int minPrecision = getMinPrecision(this.precisionMathContext.getPrecision(),
+				subtrahend.getMathContext().getPrecision());
+		MathContext resultMathContext = new MathContext(minPrecision, DEFAULT_PRECISION_ROUNDING_MODE);
```
ctjoreilly, line 268: I think another method getResultMathContext() should be written to reduce the repeated code above.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -0,0 +1,381 @@
...
+	 * 
+	 * @param exponent
+	 *            of integer type.
+	 * 
+	 * @result a new BigDecimalProbabilityNumber that is this
+	 *         BigDecimalProbabilityNumber raised to the exponent value.
+	 */
+	@Override
+	public ProbabilityNumber pow(int exponent) {
+		return new BigDecimalProbabilityNumber(value.pow(exponent, precisionMathContext));
```
ctjoreilly, line 325: what happens if exponent > EXPONENT_MAX, you should handle this here and delegate to the BigInteger version.

extra/src/main/java/aima/extra/probability/ProbabilityFactory.java
```diff
@@ -0,0 +1,91 @@
...
+import java.math.RoundingMode;
+
+/**
+ * Factory class to generate ProbabilityNumber types.
+ * 
+ * @author Nagaraj Poti
+ *
+ */
+public class ProbabilityFactory {
+
```
ctjoreilly, line 15: Not sure this is how I'd approach the factory mechanism but lets experiment with for now.

test/src/test/java/aima/test/unit/probability/BigDecimalTypeTest.java
```diff
@@ -0,0 +1,112 @@
+package aima.test.unit.probability;
+
+import aima.extra.probability.*;
+import static org.junit.Assert.*;
+import java.math.*;
+
+import org.junit.Test;
+
+public class BigDecimalTypeTest {
+
```
ctjoreilly, line 10: NOTE: no tests should output to system.out (i.e. if all tests successfully no output should go to the console).
Also, it would be nice to see a common set of probability tests that work irrespective of the underlying implementation (like some of the parameterized search tests, e.g.: UniformCostSearchTest.

ctjoreilly (approved):
Getting close. Will merge as is but would like you to take review feedback into consideration as part of your next set of changes.

Thanks.

PR #321, opened 2017-06-22
Update ProbabilityNumber interface and implementations

ngrj93:
Made changes that were mentioned in the previous pull request as well added code corresponding to last discussion. 

Added a method to check whether the value held by the ProbabilityNumber is valid or not (between 0 and 1). Improper computations may cause the value to exceed these bounds. If tight checks on these bounds are placed when performing a computation, intermediate computations would need to be arranged such that the result of any intermediate computation in the sequence always results in a valid probability value.

### DIFF
--- a/extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
+++ b/extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
@@ -8,16 +8,16 @@ import java.math.RoundingMode;
 /**
  * A simple wrapper class for operations with the primitive Java double
  * datatype. The maximum precision achievable is that supported by double
- * datatype. The double primitive type conforms to the IEEE 754 standard's 64
- * bit double precision format (1 sign bit, 11 bits for the exponent and 52 bits
- * for the significand). The class is immutable.
+ * datatype (approximately 15.95 digits). The double primitive type conforms to
+ * the IEEE 754 standard's 64 bit double precision format (1 sign bit, 11 bits
+ * for the exponent and 52 bits for the significand). The class is immutable.
  * 
  * @author Nagaraj Poti
  * 
  */
-public class DoubleProbabilityNumber implements ProbabilityNumber, Comparable<ProbabilityNumber> {
+public class DoubleProbabilityNumber implements ProbabilityNumber {
 
-	// Constants
+	// Static members
 
 	/**
 	 * Default threshold for checking rounding errors.
@@ -29,23 +29,17 @@ public class DoubleProbabilityNumber implements ProbabilityNumber, Comparable<Pr
 	 * According to the IEEE 754 format, double values have a precision of 15.95
 	 * decimal digits. Here, max_precision is set to 15 digits by default.
 	 */
-	private static final Integer DEFAULT_MAX_PRECISION = MathContext.DECIMAL64.getPrecision() - 1;
+	private static Integer MAX_PRECISION = MathContext.DECIMAL64.getPrecision() - 1;
 
 	/**
 	 * RoundingMode.HALF_EVEN statistically minimizes cumulative error when
 	 * applied repeatedly over a sequence of calculations.
 	 */
-	private static final RoundingMode DEFAULT_PRECISION_ROUNDING_MODE = RoundingMode.HALF_EVEN;
+	private static RoundingMode ROUNDING_MODE = RoundingMode.HALF_EVEN;
 
 	// Internal fields
 
 	private Double value;
-
-	/**
-	 * MathContext of value.
-	 */
-	private static MathContext precisionMathContext = new MathContext(DEFAULT_MAX_PRECISION,
-			DEFAULT_PRECISION_ROUNDING_MODE);
 
 	// Constructors
 
@@ -57,8 +51,7 @@ public class DoubleProbabilityNumber implements ProbabilityNumber, Comparable<Pr
 	 */
 	public DoubleProbabilityNumber(double value) {
 		if (value < 0 || value > 1) {
-			// Probability value must between 0 and 1
-			throw new IllegalArgumentException();
+			throw new IllegalArgumentException("Probability value must be between 0 and 1");
 		}
 		this.value = value;
 	}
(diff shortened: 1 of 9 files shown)


extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
```diff
@@ -57,8 +51,7 @@
 	 */
 	public DoubleProbabilityNumber(double value) {
 		if (value < 0 || value > 1) {
-			// Probability value must between 0 and 1
-			throw new IllegalArgumentException();
+			throw new IllegalArgumentException("Probability value must be between 0 and 1");
```
ctjoreilly, line 54: ...between 0 and 1 inclusive

or

in the interval [0,1]

extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
```diff
@@ -72,70 +65,31 @@ public DoubleProbabilityNumber(double value) {
 	 */
 	public DoubleProbabilityNumber(BigDecimal value) {
 		if (null == value || value.compareTo(new BigDecimal(0)) == -1 || value.compareTo(new BigDecimal(1)) == 1) {
```
ctjoreilly, line 67: if (null == value || value.compareTo(BigDecimal.ZERO) == -1 || value.compareTo(BigDecimal.ONE) == 1) {

is more efficient/cleaner.

extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
```diff
@@ -72,70 +65,31 @@ public DoubleProbabilityNumber(double value) {
...
 	/**
+	 * The BigDecimal value returned represents the double value represented by
+	 * this class with MAX_PRECISION.
+	 * 
 	 * @return value of BigDecimal type.
 	 */
 	@Override
 	public BigDecimal getValue() {
-		return new BigDecimal(value, precisionMathContext);
+		return new BigDecimal(this.value, this.getMathContext());
```
ctjoreilly, line 83: Could keep this as an attribute instead of re-instantiating on each call.

extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
```diff
@@ -72,70 +65,31 @@ public DoubleProbabilityNumber(double value) {
...
 	/**
-	 * @return precisionMathContext set to DECIMAL64 - 1 (15 digits) precision
-	 *         and HALF_EVEN RoundingMode by default.
+	 * @return MathContext set to DECIMAL64 - 1 (15 digits) precision and
+	 *         HALF_EVEN RoundingMode by default.
 	 */
 	@Override
 	public MathContext getMathContext() {
-		return precisionMathContext;
+		return new MathContext(MAX_PRECISION, ROUNDING_MODE);
```
ctjoreilly, line 92: Could keep this as a static attribute instead of re-instantiating on each call.

extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
```diff
@@ -309,7 +244,96 @@ public boolean sumsToOne(Iterable<ProbabilityNumber> allProbabilities) {
...
+	 * 
+	 * @param mc
+	 */
+	@Override
+	public void overrideComputationPrecisionGlobally(MathContext mc) {
+		if (mc.getPrecision() > 15) {
+			throw new IllegalArgumentException("Maximum precision possible for DoubleProbabilityNumber is 15");
+		}
+		MAX_PRECISION = mc.getPrecision();
+		ROUNDING_MODE = mc.getRoundingMode();
```
ctjoreilly, line 259: This override functionality should be instance based, not class based.  The idea is that the factory/instantiation mechanism put in place will be a good place to override the default to use for all instantiated objects belonging to a particular instance creator.

extra/src/main/java/aima/extra/probability/DoubleProbabilityNumber.java
```diff
@@ -309,7 +244,96 @@ public boolean sumsToOne(Iterable<ProbabilityNumber> allProbabilities) {
...
+	 * 
+	 * @param that
+	 *            the ProbabilityNumber type that is to be compared with this
+	 *            DoubleProbabilityNumber.
+	 *
+	 * @return true if this == that, false otherwise
+	 */
+	@Override
+	public boolean equals(Object that) {
+		DoubleProbabilityNumber second = toInternalType((ProbabilityNumber) that);
```
ctjoreilly, line 274: You may want to do a conditional check on the type here:

if (that instanceof ProbabilityNumber)
...
} else {
return false;

extra/src/main/java/aima/extra/probability/LogProbabilityNumber.java
```diff
@@ -28,24 +28,18 @@
...
 	 */
-	private static final Integer DEFAULT_MAX_PRECISION = MathContext.DECIMAL64.getPrecision() - 1;
+	private static Integer MAX_PRECISION = MathContext.DECIMAL64.getPrecision() - 1;
 
 	/**
 	 * RoundingMode.HALF_EVEN statistically minimizes cumulative error when
 	 * applied repeatedly over a sequence of calculations.
 	 */
-	private static final RoundingMode DEFAULT_PRECISION_ROUNDING_MODE = RoundingMode.HALF_EVEN;
+	private static RoundingMode ROUNDING_MODE = RoundingMode.HALF_EVEN;
```
ctjoreilly, line 38: As with DoubleProbabilityNumber I think these two statics may need to become instance based.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -146,4 +171,12 @@
...
 	 */
 	boolean sumsToOne(Iterable<ProbabilityNumber> allProbabilities);
+
+	/**
+	 * Override the precision of ProbabilityNumber instances returned as a
+	 * result of performing operations.
+	 * 
+	 * @param mc
+	 */
+	void overrideComputationPrecisionGlobally(MathContext mc);
```
ctjoreilly, line 181: I think this might better belong in the instantiation mechanism exclusively. Not sure.

ctjoreilly (approved):
Approving changes. Note several of the comments applied to DoubleProbabilityNumber are applicable to other variants as well. 

Two main things to decide upon:
1. override precision mechanism is currently global, i.e. static based, I think this will be more useful/flexible if instance based and handled through factory mechanism.

2. I take your point as regards strict bounds checking possibly causing issues when computing intermediate results. Perhaps we can, similar to override precision, that allows strict bound checking to be turned off during instantiation (i.e. via factory method as well and corresponding constructors take a flag).

PR #322, opened 2017-06-23
Add Probability Domain Interfaces and Implementations

ngrj93:
These are the Domain interface and implementation hierarchies. The hierarchy is similar to that of the AIMA3e model with some changes.

### DIFF
--- a/extra/src/main/java/aima/extra/probability/domain/AbstractFiniteDomain.java
+++ b/extra/src/main/java/aima/extra/probability/domain/AbstractFiniteDomain.java
@@ -1,0 +1,95 @@
+package aima.extra.probability.domain;
+
+import aima.extra.util.datastructure.BiMap;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Abstract base class of FiniteDomain implementations.
+ * 
+ * @author Nagaraj Poti
+ */
+public abstract class AbstractFiniteDomain implements FiniteDomain {
+
+	// Class members
+
+	/**
+	 * Mapping of domain values to their respective indices in subclass
+	 * datastructures.
+	 */
+	private BiMap<Object, Integer> valueToIdx = new BiMap<Object, Integer>();
+
+	// PUBLIC
+
+	// DiscreteDomain methods
+
+	@Override
+	public boolean isFinite() {
+		return true;
+	}
+
+	@Override
+	public boolean isInfinite() {
+		return false;
+	}
+
+	@Override
+	public abstract int size();
+
+	@Override
+	public abstract boolean isOrdered();
+
+	// FiniteDomain methods
+
+	@Override
+	public abstract Set<?> getPossibleValues();
+
+	@Override
+	public int getOffset(Object value) {
+		Integer idx = valueToIdx.getForward(value);
+		if (null == idx) {
+			throw new IllegalArgumentException("Value [" + value + "] is not a legal value for this domain.");
+		}
+		return idx.intValue();
+	}
+
+	@Override
+	public Object getValueAt(int offset) {
+		return valueToIdx.getBackward(offset);
+	}
+
+	/**
+	 * Check for equality of AbstractFiniteDomain instances. Two domains are
+	 * equal if they have the same values (and are in the same order if they are
+	 * both ordered). Domains that have the same values and are both ordered,
+	 * but with different orderings specified, are different.
+	 * 
+	 * @return true if both domains are equal, false otherwise.
+	 */
+	@Override
+	public boolean equals(Object o) {
+		if (!(o instanceof AbstractFiniteDomain)) {
+			return false;
+		}
+		AbstractFiniteDomain other = (AbstractFiniteDomain) o;
+		if (this.isOrdered() != other.isOrdered()) {
+			return false;
+		} else if (this.isOrdered() == other.isOrdered() && this.isOrdered() == false) {
+			return this.getPossibleValues().equals(other.getPossibleValues());
+		} else {
+			return this.valueToIdx.equals(other.valueToIdx);
+		}
+	}
+	
+	@Override
+	public int hashCode() {
+		return this.getPossibleValues().hashCode();
+	}
+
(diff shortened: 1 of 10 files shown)


extra/src/main/java/aima/extra/probability/domain/AbstractFiniteDomain.java
```diff
@@ -0,0 +1,95 @@
...
+	@Override
+	public int hashCode() {
+		return this.getPossibleValues().hashCode();
+	}
+
+	// PROTECTED
+
+	protected void indexPossibleValues(Set<?> possibleValues) {
+		final AtomicInteger count = new AtomicInteger(0);
+		possibleValues.forEach((value) -> valueToIdx.put(value, count.incrementAndGet()));
```
ctjoreilly, line 93: Won't this cause your index to start from 1 and not 0, i.e.:

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html#incrementAndGet--

?

Perhaps a better way is to avoid using the AtomicInteger altogether and do this instead:

possibleValues.forEac(value -> valueToIdx.put(value, valueToIdx.size());

extra/src/main/java/aima/extra/probability/domain/AbstractFiniteDomain.java
```diff
@@ -0,0 +1,95 @@
+package aima.extra.probability.domain;
+
+import aima.extra.util.datastructure.BiMap;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Abstract base class of FiniteDomain implementations.
+ * 
+ * @author Nagaraj Poti
```
ctjoreilly, line 10: As a rule of thumb if you are basing your code on pre-existing code you should keep the other author(s) in the list of authors. Order the authors based on who you believe contributed the most to the current design of the class - in this case you can list yourself first.

Note: I am guessing you just missed here as you have done this in the other code files.

extra/src/main/java/aima/extra/probability/domain/FiniteArbitraryTokenDomain.java
```diff
@@ -0,0 +1,67 @@
...
+ * 
+ * @author Ciaran O'Reilly
+ * @author Nagaraj Poti
+ */
+public class FiniteArbitraryTokenDomain extends AbstractFiniteDomain {
+
+	/**
+	 * Set of all possible values for the domain.
+	 */
+	private Set<Object> possibleValues = null;
```
ctjoreilly, line 21: Should we use Set<?> instead of Set<Object> here?

extra/src/main/java/aima/extra/probability/domain/FiniteBooleanDomain.java
```diff
@@ -0,0 +1,66 @@
...
+	 * Single instance of FiniteBooleanDomain.
+	 */
+	private static FiniteBooleanDomain onlyInstance = null;
+
+	/**
+	 * Set of possible values for boolean domain consists of {true, false}.
+	 * Consistent order is ensured.
+	 */
+	private Set<Boolean> possibleValues = new LinkedHashSet<Boolean>(Arrays.asList(Boolean.TRUE, Boolean.FALSE));
+
```
ctjoreilly, line 28: possibleValues in the case can be static as it will never change. Additionally, if you use:

Collections.unmodifiableSet(new Linked)

it saves you instantiation an unmodifiable set wrapper every time you call getPossibleValues(). Ideaily, this pattern should be followed where applicable in all the domain implementations as it is more efficiet.

extra/src/main/java/aima/extra/probability/domain/FiniteBooleanDomain.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	@Override
+	public boolean isOrdered() {
+		return false;
+	}
+
+	@Override
+	public Set<Boolean> getPossibleValues() {
+		return Collections.unmodifiableSet(possibleValues);
```
ctjoreilly, line 64: As mentioned above don't like doing this here as it introduces unneeded overhead.

extra/src/main/java/aima/extra/probability/domain/FiniteDomain.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Set;
+
+/**
+ * A Domain over a countable/discrete and finite set of objects.
+ * 
+ * @author Ciaran O'Reilly
+ * 
+ */
+public interface FiniteDomain extends DiscreteDomain {
+	/**
```
ctjoreilly, line 12: should we add generics to the domain hierarchy for the types of their values? May be too restrictive.

extra/src/main/java/aima/extra/util/datastructure/BiMap.java
```diff
@@ -0,0 +1,66 @@
...
+ * Bidirectional map datastructure.
+ *
+ * @param <K>
+ *            Key in the forward direction, value in the backward direction.
+ * @param <V>
+ *            Value in the forward direction, key in the backward direction.
+ * 
+ * @author Nagaraj Poti
+ */
+public class BiMap<K extends Object, V extends Object> {
```
ctjoreilly, line 16: Not sure you require the extends Object here in the generics declaration.

ctjoreilly (approved):
Minor feedback that I'd like you to address. Will merge to trunk as is.

extra/src/main/java/aima/extra/probability/domain/AbstractFiniteDomain.java
ngrj93: The name is missing in the original file, hence didn't put it there.

extra/src/main/java/aima/extra/probability/domain/FiniteDomain.java
ngrj93: I was thinking the same of adding generics from the higher interface. It is unlikely that different object types would be used to formulate the domain.

PR #323, opened 2017-06-26
Add ProbTensor and update ProbabilityNumber tests

ngrj93:
- Added ProbTensor class responsible for performing probability arithmetic by overriding intermediary result validity checks and allows for arbitrary precision to be specified if needed.

- Updated the tests section

### DIFF
--- a/extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
+++ b/extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
@@ -14,7 +14,7 @@ import java.math.RoundingMode;
  * @author Nagaraj Poti
  *
  */
-public class BigDecimalProbabilityNumber implements ProbabilityNumber {
+public class BigDecimalProbabilityNumber extends ProbabilityNumber {
 
 	// Static members
 
@@ -25,11 +25,9 @@ public class BigDecimalProbabilityNumber implements ProbabilityNumber {
 	private static final int EXPONENT_MAX = 999999999;
 
 	/**
-	 * Global precision setting for operations. Set to null by default but can
-	 * be overriden to ensure all computation results are specified to this
-	 * precision.
+	 * Precision value corresponding to MathContext.UNLIMITED.
 	 */
-	private static MathContext ARBITRARY_PRECISION = null;
+	private static final Integer UNLIMITED_PRECISION = 0;
 
 	/**
 	 * Constants of UNLIMITED precision.
@@ -59,10 +57,17 @@ public class BigDecimalProbabilityNumber implements ProbabilityNumber {
 	 *            to be assigned.
 	 */
 	public BigDecimalProbabilityNumber(double value) {
-		if (value < 0 || value > 1) {
-			throw new IllegalArgumentException("Probability value must be between 0 and 1");
-		}
-		this.value = BigDecimal.valueOf(value);
+		this(BigDecimal.valueOf(value), null);
+	}
+
+	/**
+	 * Construct a BigDecimalProbabilityNumber from BigDecimal type.
+	 * 
+	 * @param value
+	 *            to be assigned.
+	 */
+	public BigDecimalProbabilityNumber(BigDecimal value) {
+		this(value, null);
 	}
 
 	/**
@@ -75,24 +80,18 @@ public class BigDecimalProbabilityNumber implements ProbabilityNumber {
 	 *            MathContext to be associated with value.
 	 */
 	public BigDecimalProbabilityNumber(BigDecimal value, MathContext mc) {
-		if (value.compareTo(BIG_DECIMAL_ZERO) == -1 || value.compareTo(BIG_DECIMAL_ONE) == 1) {
-			throw new IllegalArgumentException("Probability value must be between 0 and 1");
-		} else if (null == mc) {
-			this.value = value;
-		} else {
+		if (null == value) {
+			throw new IllegalArgumentException("A probability value must be specified.");
+		}
+		if (checkRequired) {
+			checkValidityOfArguments(value);
+		}
+		if (null != mc) {
 			this.value = new BigDecimal(value.toString(), mc);
 			this.roundingMode = mc.getRoundingMode();
+		} else {
+			this.value = value;
 		}
-	}
-
-	/**
-	 * Construct a BigDecimalProbabilityNumber from BigDecimal type.
-	 * 
-	 * @param value
-	 *            to be assigned.
-	 */
-	public BigDecimalProbabilityNumber(BigDecimal value) {
-		this(value, null);
 	}
 
 	// Public methods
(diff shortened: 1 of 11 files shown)


extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -5,29 +5,38 @@
...
  * 
- * The interface is intended to be immutable. Implementations may implement
+ * The subclasses are intended to be immutable. Implementations may implement
  * overriden versions of equals and hashcode methods.
  * 
  * @author Nagaraj Poti
  *
  */
-public interface ProbabilityNumber extends Comparable<ProbabilityNumber> {
+public abstract class ProbabilityNumber implements Comparable<ProbabilityNumber> {
```
ctjoreilly, line 23: Why the switch to an abstract class? Would it not be better to leave as an interface and add an explicit AbstractProbabilityNumber class for use by concrete implementations where appropriate.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -5,29 +5,38 @@
...
+
+	/**
+	 * checkRequired is true (by default) if ProbabilityNumber should be checked
+	 * for validity, false if checks are to be disabled (required for performing
+	 * a sequence of computations where the intermediary results may temporarily
+	 * lie outside the boundaries of a probability value).
+	 */
+	public static boolean checkRequired = true;
```
ctjoreilly, line 31: This should not be a static setting as it will act globally, which is likely not what you want.

extra/src/main/java/aima/extra/probability/ProbabilityNumber.java
```diff
@@ -140,7 +163,21 @@
...
+	public abstract ProbabilityNumber divide(ProbabilityNumber divisor);
+	
+	/**
+	 * Divide two ProbabilityNumber types and return a ProbabilityNumber
+	 * representing the result of division.
+	 * 
+	 * @param divisor
+	 * 
+	 * @param mc
+	 *            MathContext for ARBITRARY_PRECISION computation.
```
ctjoreilly, line 175: Is DEFINED_PRECISION not a better more accurate name than ARBITRARY_PRECISION?

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -25,11 +25,9 @@
 	private static final int EXPONENT_MAX = 999999999;
 
 	/**
-	 * Global precision setting for operations. Set to null by default but can
-	 * be overriden to ensure all computation results are specified to this
-	 * precision.
+	 * Precision value corresponding to MathContext.UNLIMITED.
 	 */
-	private static MathContext ARBITRARY_PRECISION = null;
+	private static final Integer UNLIMITED_PRECISION = 0;
```
ctjoreilly, line 30: Why not just use MathContext.UNLIMITED?

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -59,40 +57,41 @@
...
+	public BigDecimalProbabilityNumber(BigDecimal value, MathContext mc) {
+		if (null == value) {
+			throw new IllegalArgumentException("A probability value must be specified.");
+		}
+		if (checkRequired) {
+			checkValidityOfArguments(value);
+		}
+		if (null != mc) {
+			this.value = new BigDecimal(value.toString(), mc);
+			this.roundingMode = mc.getRoundingMode();
```
ctjoreilly, line 91: toString() can be very inefficient, in particular with large numbers. Investigate using this constructor instead:
BigDecimal(BigInteger unscaledVal, int scale, MathContext mc)
the unscaledVal and scale values are accessible on the BigDecimal API.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -195,8 +194,31 @@ public ProbabilityNumber subtract(ProbabilityNumber that) {
...
+	public ProbabilityNumber multiply(ProbabilityNumber that, MathContext mc) {
 		BigDecimalProbabilityNumber multiplier = toInternalType(that);
-		MathContext resultMathContext = getResultMathContext(this.getMathContext(), multiplier.getMathContext());
+		MathContext resultMathContext;
+		if (null != mc) {
+			resultMathContext = mc;
+		} else {
+			resultMathContext = getResultMathContext(this.getMathContext(), multiplier.getMathContext());
+		}
 		return new BigDecimalProbabilityNumber(this.value.multiply(multiplier.value, resultMathContext));
```
ctjoreilly, line 222: This is just advice for writing code that is easier to debug. Instead of doing:

return new ...

always assign a local variable which you return, e.g."

result = new ...
return result;

This lets you easily inspect the return value in a debugger before stepping out of the function.

extra/src/main/java/aima/extra/probability/BigDecimalProbabilityNumber.java
```diff
@@ -340,6 +420,19 @@ public String toString() {
...
+	 * 
+	 * @param value
+	 *            to be assigned to BigDecimalProbabilityNumber value.
+	 */
+	public static void checkValidityOfArguments(BigDecimal value) {
+		if (value.compareTo(BIG_DECIMAL_ZERO) == -1 || value.compareTo(BIG_DECIMAL_ONE) == 1) {
+			throw new IllegalArgumentException("Probability value must be in the interval [0,1].");
+		}
+	}
+
```
ctjoreilly, line 435: Good.

extra/src/main/java/aima/extra/probability/ProbTensor.java
```diff
@@ -0,0 +1,153 @@
...
+	/**
+	 * Add two ProbabilityNumber operands and return the result.
+	 * 
+	 * @param operand1
+	 * @param operand2
+	 * 
+	 * @return ProbabilityNumber
+	 */
+	public ProbabilityNumber add(ProbabilityNumber operand1, ProbabilityNumber operand2) {
+		synchronized (ProbabilityNumber.class) {
```
ctjoreilly, line 28: deadlock deadlock deadlock!!!!! this class based approach is not a good idea.

ctjoreilly: I recommend adding a class instance attribute that determines if checks are to be performed and extend the ProbabilityNumber API to accept an interval check flag as well as the existing math context flag.

extra/src/main/java/aima/extra/probability/ProbTensor.java
```diff
@@ -0,0 +1,153 @@
...
+	}
+	
+	/**
+	 * Override the precision of ProbabilityNumber instances returned as a
+	 * result of performing operations.
+	 * 
+	 * @param mc
+	 */
+	public void overrideComputationPrecision(MathContext mc) {
+		this.ARBITRARY_PRECISION = mc;
```
ctjoreilly, line 150: bad name. I would used DEFINED_PRECISION or OVERRIDE_PRECISION.

ctjoreilly (changes requested):
Would prefer ProbabilityNumber remain an interface and introduce an AbstractProbabilityNumber class to achieve shared functionality where implemented. Like the idea of the ProbabilityTensor helper class but not the static based implementation - an instance based approach needs to be taken here.

PR #325, opened 2017-07-11
Update ProbabilityFactory, add ProbabilityComputation

ngrj93:


core/src/main/java/aima/core/util/math/MixedRadixInterval.java
```diff
@@ -110,6 +110,38 @@ public BigInteger getValueFor(int[] numeralValues) {
...
+	 *            when computing the mixed interval value.
+	 *
+	 * @return the value of the numerals at within the interval.
+	 */
+	public BigInteger subsetInterval(int numeralValues[], int leaveOutSubset[]) {
+		Arrays.sort(leaveOutSubset);
+		BigInteger cvalue = BigInteger.ZERO;
+		BigInteger mvalue = BigInteger.ONE;
+		int c = -1, j = leaveOutSubset.length - 1;
+		for (int i = numeralValues.length - 1; i >= 0; i--) {
```
ctjoreilly: c and j index values are not updated in following logic.
can you add test cases for this method so it will be easier to follow its functionality.

extra/src/main/java/aima/extra/probability/constructs/ProbabilityComputation.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * ProbabilityComputation is a helper class which allows for ProbabilityNumber
+ * operations to return ProbabilityNumber objects that may contain invalid
+ * values. It is possible for intermediate calculations to result in invalid
+ * probability values, whereas the outcome after applying all the operations
+ * results in a valid probability value.
+ * 
+ * @author Nagaraj Poti
+ */
+public class ProbabilityComputation {
```
ctjoreilly: Remove checkrequired logic as it makes things complicated and introduces side effects and potential concurrency issues in multi-threaded environments. 

Update class javadoc to indicate use of this class in case of streams as this is the primary intent of this class.

ctjoreilly (changes requested):
Initial feedback as discussed on hangout.

ngrj93:
Performed changes  -> 
1. Removed checkRequired logic
2. Removed subsetInterval logic in MixedRadixInterval (rewrote logic for ProbabilityTable section)

PR #328, opened 2017-07-27
Add initial bayes implementation

ngrj93:
Will be updating the same commit over the course of the day.


extra/src/main/java/aima/extra/probability/bayes/AbstractNode.java
```diff
@@ -0,0 +1,64 @@
...
+
+	private List<Node> parents;
+
+	private List<Node> children;
+
+	public AbstractNode(RandomVariable var) {
+		this(var, null);
+	}
+
+	public AbstractNode(RandomVariable var, List<? extends Node> parents) {
```
ctjoreilly, line 22: do we need to use <? extends Node> here?

extra/src/main/java/aima/extra/probability/bayes/AbstractNode.java
```diff
@@ -0,0 +1,64 @@
...
+		this.nodeVariable = var;
+		parents.stream().forEach(e -> ((AbstractNode) e).addChild(this));
+		this.parents = parents.stream().collect(Collectors.toCollection(ArrayList<Node>::new));
```
ctjoreilly, line 25: would: this.parents = new ArrayList<>(parents); not be simpler?

extra/src/main/java/aima/extra/probability/bayes/AbstractNode.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	@Override
+	public RandomVariable getNodeVariable() {
+		return this.nodeVariable;
```
ctjoreilly, line 30: 'this.' not required here.

extra/src/main/java/aima/extra/probability/bayes/AbstractNode.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	@Override
+	public List<Node> getMarkovBlanket() {
+		return null;
+	}
+
+	@Override
+	public ConditionalProbabilityDistribution getCPD() {
+		// TODO Auto-generated method stub
```
ctjoreilly, line 56: // TODO - implement

extra/src/main/java/aima/extra/probability/bayes/AbstractNode.java
```diff
@@ -0,0 +1,64 @@
...
+		return null;
+	}
+
+	protected void addChild(Node childNode) {
+
```
ctjoreilly, line 61: // TODO - implement

extra/src/main/java/aima/extra/probability/bayes/AbstractProbabilityTable.java
```diff
@@ -0,0 +1,217 @@
...
+					Collections.nCopies(expectedSize, this.probFactory.valueOf(BigDecimal.ZERO)));
+		} else {
+			if (values.size() != expectedSize) {
+				throw new IllegalArgumentException(
+						"ProbabilityTable of length " + values.size() + " is not the correct size, should be "
+								+ expectedSize + " in order to represent all possible combinations.");
+			}
+			this.values = values.stream().map(val -> this.probFactory.convert(val)).collect(Collectors.toList());
+		}
+		this.randomVariables = vars.stream().map(var -> (RandomVariable) var).collect(Collectors.toList());
```
ctjoreilly, line 92: Again, why not just instantiate the Array<>(vars) directly instead of using the stream mechanism, which is more verbose and likely slower.

extra/src/main/java/aima/extra/probability/bayes/AbstractProbabilityTable.java
```diff
@@ -0,0 +1,217 @@
...
+	/**
+	 * Contains all values of the distribution.
+	 */
+	protected List<ProbabilityNumber> values;
+
+	/**
+	 * Ordered set of all random variables over which the distribution is on.
+	 */
+	protected List<RandomVariable> randomVariables;
+
```
ctjoreilly, line 40: Tables are immutable therefore you should be able to get away with using arrays (i.e. ProbabilityNumber[] and RandomVariable[] as these are more efficient for a large number of accesses, which will be the case, compared to using the List access methods).

extra/src/main/java/aima/extra/probability/bayes/AbstractProbabilityTable.java
```diff
@@ -0,0 +1,217 @@
...
+	 */
+	public AbstractProbabilityTable(List<RandomVariable> vars, List<ProbabilityNumber> values,
+			Class<? extends ProbabilityNumber> clazz) {
+		// TODO - Unmodifiable list ?
+		int expectedSize = ProbabilityUtilities.expectedSizeofProbabilityTable(vars);
+		this.clazz = clazz;
+		this.probFactory = ProbabilityFactory.make(clazz);
+		if (null == values) {
+			this.values = new ArrayList<ProbabilityNumber>(
+					Collections.nCopies(expectedSize, this.probFactory.valueOf(BigDecimal.ZERO)));
```
ctjoreilly, line 83: Not sure, this is a good idea, the table should be emutable and therefore defaulting to all zeros is meaningless.

extra/src/main/java/aima/extra/probability/bayes/AbstractProbabilityTable.java
```diff
@@ -0,0 +1,217 @@
...
+		return this.randomVariables;
+	}
+
+	@Override
+	public boolean contains(RandomVariable var) {
+		return this.randomVariables.contains(var);
+	}
+
+	@Override
+	public ProbabilityNumber getValue(Predicate<List<Object>> eventProposition) {
```
ctjoreilly, line 112: Not sure I am keen on this signature 'Predicate<List<Object>>' it should be something more like 'Predicate<Map<RandomVariable, Object>>' to assignments to random variables are explicit.

extra/src/main/java/aima/extra/probability/bayes/AbstractProbabilityTable.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import aima.core.util.math.MixedRadixInterval;
+import aima.extra.probability.ProbabilityNumber;
+import aima.extra.probability.RandomVariable;
+import aima.extra.probability.ProbabilityComputation;
+import aima.extra.probability.constructs.ProbabilityUtilities;
+import aima.extra.probability.domain.FiniteDomain;
+import aima.extra.probability.factory.ProbabilityFactory;
+import aima.extra.probability.proposition.Proposition;
```
ctjoreilly, line 20: I thought we were dropping Proposition?

extra/src/main/java/aima/extra/probability/bayes/ConditionalProbabilityTable.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * A conditional probability table, or CPT, can be used for representing
+ * conditional probabilities for discrete (finite) random variables. Each row in
+ * a CPT contains the conditional probability of each node value for a
+ * <b>conditioning case</b>.
+ * 
+ * @author Ciaran O'Reilly
+ * @author Nagaraj Poti
+ */
```
ctjoreilly, line 20: Why the switch from an interface to a class?

extra/src/main/java/aima/extra/probability/bayes/FiniteNode.java
```diff
@@ -0,0 +1,17 @@
...
+
+import aima.extra.probability.bayes.ConditionalProbabilityTable;
+
+/**
+ * A node over a Random Variable that has a finite countable domain.
+ * 
+ * @author Ciaran O'Reilly
+ */
+public interface FiniteNode {
+
```
ctjoreilly, line 11: should this not still extend Node?

extra/src/main/java/aima/extra/probability/bayes/Node.java
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * @return the random variable represented by this Node.
+	 */
+	RandomVariable getNodeVariable();
+
+	/**
+	 * @return true if this Node has no parents.
+	 */
+	boolean isRoot();
```
ctjoreilly, line 29: could implement as a default method based on getParents().size() == 0

extra/src/main/java/aima/extra/probability/bayes/ProbabilityMass.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * ProbabilityMass is the probability distribution for discrete random
+ * variables. Only a finite number of random variables is supported.
+ * 
+ * @author Ciaran O'Reilly
+ * @author Nagaraj Poti
+ */
+public interface ProbabilityMass extends ProbabilityDistribution {
+
```
ctjoreilly, line 17: this interface should not be specific to bayes.

extra/src/main/java/aima/extra/probability/proposition/Proposition.java
```diff
@@ -0,0 +1,43 @@
+package aima.extra.probability.proposition;
+
```
ctjoreilly, line 2: why are the proposition classes still around, I thought we were getting rid of these in preference to using predicates?

ctjoreilly (approved):
Please review comments and incorporate in future changes. Also, please reduce the size of your push requests. It is more efficient to have smaller more frequent cycles.

PR #329, opened 2017-08-01
Update Bayes implementation (Tables and Interfaces)

ngrj93:
Preferred list over arrays because of stream operations (would have to write some functions for arrays). Tried doing it with arrays as well, but for some generic functions, typecasting of arrays was required additionally.


extra/src/main/java/aima/extra/probability/bayes/AbstractNode.java
```diff
@@ -19,15 +18,15 @@ public AbstractNode(RandomVariable var) {
...
 	}
 
-	public AbstractNode(RandomVariable var, List<? extends Node> parents) {
-		this.nodeVariable = var;
-		parents.stream().forEach(e -> ((AbstractNode) e).addChild(this));
-		this.parents = parents.stream().collect(Collectors.toCollection(ArrayList<Node>::new));
+	public AbstractNode(RandomVariable var, List<Node> parents) {
+		nodeVariable = var;
+		parents.stream().forEach(e -> ((AbstractNode)e).addChild(this));
+		this.parents = new ArrayList<Node>(parents);
```
ctjoreilly, line 24: this.parents = new ArrayList<>(parents);

extra/src/main/java/aima/extra/probability/bayes/AbstractProbabilityTable.java
```diff
@@ -105,21 +126,16 @@ public AbstractProbabilityTable(List<RandomVariable> vars, List<ProbabilityNumbe
 
 	@Override
 	public boolean contains(RandomVariable var) {
-		return this.randomVariables.contains(var);
+		boolean contains = this.randomVariables.stream().anyMatch(x -> x.equals(var));
```
ctjoreilly, line 129: Do you think faster or slower than using contains()?

ctjoreilly (approved):

PR #330, opened 2017-08-10
Update basic ProbabilityTable tests and other fixes.

ngrj93:


extra/src/main/java/aima/extra/probability/RandVar.java
```diff
@@ -70,4 +71,17 @@ public boolean equals(Object o) {
 		boolean result = this.name.equals(other.name) && this.domain.equals(other.domain);
 		return result;
 	}
+
+	/**
+	 * Two RandVar objects have the same hashcode if they have the same names.
+	 */
+	@Override
+	public int hashCode() {
+		return Objects.hash(name);
```
ctjoreilly, line 80: Note: Objects.hash is intended to be used with a sequence of objects. For a single object as in this case just use name.hashCode

test/src/test/java/aima/test/unit/probability/DoubleTypeTest.java
```diff
@@ -8,81 +8,82 @@
 import org.junit.Test;
 
 /**
- * ProbabilityTest to check the various functions of the DoubleProbabilityNumber class
+ * ProbabilityTest to check the various functions of the DoubleProbabilityNumber
+ * class
  */
 public class DoubleTypeTest {
```
ctjoreilly, line 14: Shouldn't this be generalized to a ProbabilityNumberTest and test all the different types of probability numbers using JUnits parameterized tests methods?

ctjoreilly (approved):

PR #333, opened 2017-08-27
Add ProbabilityModel and add spliterator implementation

ngrj93:


core/src/main/java/aima/core/util/math/MixedRadixInterval.java
```diff
@@ -58,6 +58,10 @@ public MixedRadixInterval(int[] radices, int[] leftEndPointNumeralValues, int[]
 		}
 		return last;
 	}
+	
+	public int[] getRadices() {
+		return Arrays.copyOf(radices, radices.length);
+	}
```
ctjoreilly, line 65: Why the need for the copy? This could be expensive if used in the wrong place.

extra/src/main/java/aima/extra/probability/RandVar.java
```diff
@@ -32,9 +34,8 @@
 	 */
 	public RandVar(String name, Domain domain) {
 		ProbabilityUtilities.checkValidRandomVariableName(name);
-		if (null == domain) {
-			throw new IllegalArgumentException("Domain of RandVar must be specified.");
-		} else if (!(domain instanceof FiniteDomain)) {
+		Objects.requireNonNull("Domain of RandVar must be specified.");
```
ctjoreilly, line 37: I think the wrong requireNonNull method is being called as you are not actually passing the 'domain' argument to it (this is just testing if the message is not null which it will never be).

extra/src/main/java/aima/extra/probability/bayes/AbstractProbabilityTable.java
```diff
@@ -24,7 +32,7 @@
  * 
  * @author Nagaraj Poti
  */
-public abstract class AbstractProbabilityTable implements CategoricalDistribution {
+public abstract class AbstractProbabilityTable implements CategoricalDistribution, Iterable<Map<RandomVariable, Object>> {
```
ctjoreilly, line 35: makes sense.

ctjoreilly (approved):

PR #334, opened 2017-08-29
Update bayes model implementation and enumeration ask

ngrj93:
I had added a few lines to the previous commit, which I believe is causing the mergeability problem.

Will be posting tests soon.


ctjoreilly:
This looks like a mixed commit (i.e. includes stuff from your last pull request again). Can you close out and resolve?

PR #335, opened 2017-10-02
Size Specified Learner 

s-avni:
Size Specified Learner implemented (CrossValidationWrapper).
Example for restaurant decision tree created.


norvig:
Thanks for the work, but we already merged another implementation of cross validation.

PR #337, opened 2017-10-04
Implemented bidirectional search enhancement

HasinduGajanayake:
I implemented an enhancement for the  bidirectional search .I hope it will be a optimal solution


RuedigerLunde:
Sorry - this pull request does not compile. Callers of this class expect a specialization of QueueSearch but the proposed implementation ignores this requirement completely.

PR #351, opened 2018-02-28
Adds DTAgent

samagra14:
This branch adds Decision Theoretic Agent to the AIMA4e branch. I will be adding DTAgent to the AIMA3e branch. The same code cannot be applied to AIMA3e ( because of different APIs). However, I will add the implementation to the 3E branch soon.

### DIFF
--- a/core/src/main/java/aima/core/Probability/BeliefState.java
+++ b/core/src/main/java/aima/core/Probability/BeliefState.java
@@ -1,0 +1,8 @@
+package aima.core.Probability;
+
+public interface BeliefState<A,P> {
+    void update(A action,P percept);
+
+
+
+}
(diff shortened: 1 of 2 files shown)


samagra14:
@norvig 
Please review this PR. I will add DTAgent to AIMA3e as well. Besides, I am having a hard time thinking of any tests that can be applied to this abstract class. Can you please provide me with some hints regarding the same?

norvig:
You wouldn't test an abstract class - you test instantiations of it.

samagra14:
@norvig 
I am aware of the fact. I just meant that whether I should extend this abstract class into a concrete one and then test it. There are only three methods in this abstract class. I just needed to know if tests are required for this Agent.

PR #355, opened 2018-03-05
Added Testing Section and Notebook section 

JaeDukSeo:
So I saw one of the issues and decide to make this change. Also, since Jupiter notebook supports Java now, I think it is a good idea to have a short tutorial here as well. Let me know what you think.  (https://github.com/jupyter/jupyter/wiki/Jupyter-kernels)


norvig:
Having a lot of empty tests/notebooks will encourage filling in the table, I hope.

Issue #371, opened 2018-03-15
Implement Environment in AIMA4e


ritwik12:
Add Environment to AIMA4e.
Environment description from AIMA3e : 
1. Fig - 2      Page 34

PR #378, opened 2018-03-17
Add Environment.java to AIMA4e 

ritwik12:
Solves #371 This PR consists :   
1. Add Environment in README.md.
2. Add Environment.java.

### DIFF
--- a/core/src/main/java/aima/core/agent/basic/Environment.java
+++ b/core/src/main/java/aima/core/agent/basic/Environment.java
@@ -1,0 +1,33 @@
+package aima.core.agent.basic;
+
+import aima.core.agent.api.Agent;
+import java.util.List;
+
+/**
+ * Possible environments in which Agent(s) can percieve and act.
+ * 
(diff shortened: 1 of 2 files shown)


core/src/main/java/aima/core/agent/basic/Environment.java
```diff
@@ -0,0 +1,33 @@
...
+ * Possible environments in which Agent(s) can percieve and act.
+ * 
+ * @author Ritwik Sharma
+ */
+public interface Environment {
+	
+	/**
+	 * Returns the Agent(s) belonging to this Environment.
+	 */
+	List<Agent> getAgents();
```
utsavoza, line 16: Agent in `AIMA4e` is a generic interface with a different API than in `AIMA3e`. Any implementation of this interface would possibly be fraught of unchecked warnings. Please verify the implementation by giving a supporting example environment that implements this interface.

ritwik12:
@norvig Please have a look

PR #380, opened 2018-03-18
In class javadoc, replace <- and &lt;- with &larr;

okulev:
Pseudocode in class javadoc often uses `<-` for assignment. This is not proper javadoc and confuses the NetBeans editor. Sometimes `&lt;-` is used, which is valid, and sometimes `&larr;` is used.

This pull request replaces all assignment operators with `&larr;`. It is guaranteed to be proper javadoc, and also this character is in [WGL4](https://en.wikipedia.org/wiki/Windows_Glyph_List_4), so it's generally available everywhere.

There are also several instances of `&harr;` replacement.

The instances of `<-` that are not in pseudocode javadoc are not changed.

The replacement was done with Search and Replace regexps in files and then manually verified afterwards.


norvig:
Code now has literal ← character, which is better than `&larr;`

PR #384, opened 2018-03-20
Fix incorrect method call argument

utsavoza:


utsavoza:
@norvig @RuedigerLunde  The current method intends to remove the mapping of the key pair from the map, however, it incorrectly removes the `key2` by passing the map i.e. `k1map` as an argument in `remove` method instead of `key2`.

Issue #188, opened 2016-03-22
AIMA4e reimplement Chp 17 Markov Decision Process from AIMA3e branch


ctjoreilly:

TheTour1st:
Hi, I have gone through the Markov Decision Process implementation of AIMA3e branch and would like work on this issue.

TheTour1st:
I was kind of hoping that you could tell me what is expected in the mentioned re-implementation.

ritwik12:
I am up for this

PR #386, opened 2018-03-21
Add Markov Decision Process to AIMA4e

ritwik12:
Solves #188.
This PR consists:
1. Readme.md edit to add Markov Decision Process
2. Add MarkovDecisionProcess.java - reimplemented for AIMA4e and improved java doc to make it more closer according to the third edition AIMA book page 647 chapter 17.

### DIFF
--- a/core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
+++ b/core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
@@ -52,7 +52,7 @@ public interface MarkovDecisionProcess<Stype, Atype> {
 	 *            the state.
 	 * @return the set of actions for state s.
 	 */
-	Set<Atype> ACTIONS(Stype s);
+	Set<Atype> Actions(Stype s);
 
 	/**
 	 * Return the probability of going from state s using action a to s' based
--- a/core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
+++ b/core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
@@ -1,0 +1,83 @@
+package aima.core.Probability;
+
+import java.util.Set;
+
+/**
+ * Artificial Intelligence A Modern Approach (4th Edition): Figure ?.?, page
+ * ???.<br>
+ * <br>
+ * 
+ * A sequential decision problem for a fully observable, stochastic environment
+ * with a Markovian transition model and additive rewards is called a <b>Markov
+ * decision process</b>, or <b>MDP</b>, and consists of a set of states (with an
+ * initial state s<sub>0</sub>; a set ACTIONS(s) of actions in each state; a
+ * transition model P(s' | s, a); and a reward function R(s).<br>
+ * <br>
+ * <b>Note:</b> Some definitions of MDPs allow the reward to depend on the
+ * action and outcome too, so the reward function is R(s, a, s'). This
+ * simplifies the description of some environments but does not change the
+ * problem in any fundamental way.
+ * 
+ * @param <Stype>
+ *            the state type.
+ * @param <Atype>
+ *            the action type.
+ * 
+ * @author Ritwik Sharma
+ * @author Ciaran O'Reilly
+ * @author Ravi Mohan
+ * 
+ */
+public interface MarkovDecisionProcess<Stype, Atype> {
+
+	/**
+	 * Get the set of states associated with the Markov decision process.
+	 * 
+	 * @return the set of states associated with the Markov decision process.
+	 */
+	Set<Stype> states();
+
+	/**
+	 * Get the initial state s<sub>0</sub> for this instance of a Markov
+	 * decision process.
+	 * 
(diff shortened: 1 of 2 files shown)


utsavoza (changes requested):
Please do not simply copy the implementations of one branch to another. The currently implemented API is different on both the branches.

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
```diff
@@ -0,0 +1,84 @@
...
+ *            the state type.
+ * @param <Atype>
+ *            the action type.
+ * 
+ * @author Ritwik Sharma
+ * @author Ciaran O'Reilly
+ * @author Ravi Mohan
+ * 
+ */
+public interface MarkovDecisionProcess<Stype, Atype extends Agent> {
```
utsavoza: I am not sure how an action type extends an Agent?

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
```diff
@@ -0,0 +1,84 @@
+package aima.core.Probability;
+
+import aima.core.agent.api.Agent;
+
+import java.util.Set;
+
+/**
+ * Artificial Intelligence A Modern Approach (3rd Edition): page 647.<br>
```
utsavoza: You are proposing changes to `AIMA4e` branch, the javadoc should reflect the change.

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
```diff
@@ -0,0 +1,84 @@
...
+	Stype getInitialState();
+
+	/**
+	 * Get the set of actions for state s.
+	 * 
+	 * @param s
+	 *            the state.
+	 * @return the set of actions for state s.
+	 */
+	Set<Atype> ACTIONS(Stype s);
```
utsavoza: Please be consistent with existing naming conventions used throughtout the project.

ritwik12:
@utsavoza Those are abstract methods and they work same for both the branches. Also, I have used Agent in place of Action here. and changed the java doc according to the book. you are not making any suggestions here. but simply arguing.

ritwik12:
@utsavoza and it's called a pull request because it's a "request", I am not pushing to merge this. if it's wrong it will not be merged.

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
ritwik12: here you are referring to ACTIONS?

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
ritwik12: Corrected, thanks for the correction.

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
ritwik12: The ACTIONS name is according to the method name as mentioned in the book to make it more close to the pseudo code.

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
ritwik12: we are not having Action in AIMA4e just like AIMA3e branch, for that, we need to create a new file or we could use the existing API. the best-suited solution was using Agent API of AIMA4e which does the work of Action.

utsavoza: I would suggest you provide an implementation of the interface to justify the usage of Agent type.

samagra14: I guess @utsavoza is right. And besides, I have mentioned earlier too that the APIs in both the branches are entirely different. Also, in the AIMA4e we use generics to represent actions.

ritwik12: Thanks for your reviews @utsavoza  and @samagra14. I am still learning a lot of things. Can you guys suggest what can we use in place of Agent here to make it more compatible with AIMA4e?

ritwik12: I think we don't need Agent for its function and we can implement it like this to prevent confusion.

ritwik12:
@utsavoza please close the review if it's done. Thanks

ritwik12:
@norvig  please have a look.

core/src/main/java/aima/core/Probability/MarkovDecisionProcess.java
norvig: ACTIONS should be Actions

ritwik12:
@norvig I have made the changes, please have a look now.

samagra14:
The API has changed quite a lot.and this version seems incompatible to those. Hence closing this PR.

PR #391, opened 2018-03-23
Online-DFS-Agent - Clarify the need for the additional check if (!sPrime.equals(result.get(s, a))), which is not present in the pseudocode but required to stop it looping endlessly on certain test problems - and Added unit test for TableDrivenAgent in class TableDrivenAgentTest.java .

hriaz17:
This is a PR which:

- adds a unit test for the TableDrivenAgent.

- Clarifies the need for the additional check if (!sPrime.equals(result.get(s, a))), which is not present in the pseudocode of OnlineDFSAgent but is required to stop it looping endlessly on certain test problems

 I've set up the basic structure for the unit test with a generic implementation as best as I could understand, however, I do think this needs more work. I'm also having trouble with testing non-existing sequences, so any help on that would be appreciated. @utsavoza @norvig @samagra14 please review this PR and help me improve it.

### DIFF
--- a/test/src/test/java/aima/test/unit/agent/TableDrivenAgentTest.java
+++ b/test/src/test/java/aima/test/unit/agent/TableDrivenAgentTest.java
@@ -1,17 +1,78 @@
 package aima.test.unit.agent;
 
+
+import java.util.ArrayList;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+
+import aima.core.agent.api.Agent;
+import aima.core.agent.api.Rule;
+import aima.core.agent.basic.TableDrivenAgent;
 import org.junit.Assert;
+import org.junit.Before;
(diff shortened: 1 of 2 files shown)


test/src/test/java/aima/test/unit/agent/TableDrivenAgentTest.java
```diff
@@ -1,17 +1,78 @@
...
+    public void testExistingSequences() { 
+    	
+    	Assert.fail("TODO");
+    	Assert.assertEquals(ACTION_1,
+				t1.perceive((String) percept.get(0)));
+		Assert.assertEquals(ACTION_2,
+				t1.perceive((String) percept.get(1)));
+		Assert.assertEquals(ACTION_3,
+				t1.perceive((String) percept.get(2))); 
     }
```
samagra14, line 65: Test case looks good.

samagra14 (approved):

Issue #51, opened 2015-03-12
Update implementation of - Fig 9.6 FOL-BC-ASK


GoogleCodeExporter:
```
Fig  9.6  FOL-BC-ASK  (update to more closely map to AIMA3e)


```

Original issue reported on code.google.com by `ctjoreilly@gmail.com` on 28 Nov 2009 at 12:57

ritwik12:
#395 updates only the Java doc. The implementation is not according to AIMA3e. It needs to be updated or reimplemented.

samagra14:
#410 solves this issue. Therefore, closing this issue.

PR #395, opened 2018-03-26
Update  FOL-BC-Ask Java Doc

ritwik12:
#51 Updating Java Doc according to AIMA 3rd edition. Previously Java doc was according to AIMA 2nd edition.


ritwik12:
@norvig please have a look.

samagra14 (commented):
I think the Java Doc must be updated only if the implementation is updated to match the AIMA3e version.
Currently, the implementation is not as per the 3rd edition. So the Javadoc must remain consistent with the existing code. You must re-implement the algorithm along with updating the JavaDoc changes.

ritwik12:
@samagra14 That is why I have only made changes to the text which doesn't have to do anything with implementation. I agree with you. I did this as It was not looking good due to old doc. rest is on Collaborators.

ritwik12:
@samagra14 I will try to reimplement it too.

PR #396, opened 2018-03-26
Minor improvements in the map2d package (AIMA4e)

samagra14:
This PR makes minor modifications to the map2d package. This PR makes the following changes: 
* Converts Statement Lambdas to Expression lambdas wherever required.
* Removes the unnecessary null checks
* A Map2D object was unnecessarily passed into the `getResultFunction()` and the`getGoalTestPredicate()` function. This PR safely deletes the unnecessary Map2D method parameter.


samagra14:
@norvig 
Please have a look here.

Issue #383, opened 2018-03-19
Add Implementation of Fig 29.1 POWERS-OF-2 for AIMA3e


ritwik12:

samagra14:
This issue was resolved in #397. Therefore closing this issue.

ritwik12:
@samagra14 Great

PR #397, opened 2018-03-27
implementation of PowersOf2

Medeah:
#383


ritwik12 (changes requested):
@Medeah Please add author name.

PR #399, opened 2018-03-31
update readme.md

ritwik12:
Add algorithms to readme that are yet to be implemented but are not there in readme.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,14 @@ Java implementation of algorithms from [Russell](http://www.cs.berkeley.edu/~rus
 |9|354|Demodulation|[Demodulation](/aima-core/src/main/java/aima/core/logic/fol/inference/Demodulation.java)|
 |9|354|Paramodulation|[Paramodulation](/aima-core/src/main/java/aima/core/logic/fol/inference/Paramodulation.java)|
 |9|345|Subsumption|[SubsumptionElimination](/aima-core/src/main/java/aima/core/logic/fol/SubsumptionElimination.java)|
+|9.8|341|Append|---|
+|10.1|369|AIR-CARGO-TRANSPORT-PROBLEM|---|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,7 +81,14 @@ Java implementation of algorithms from [Russell](http://www.cs.berkeley.edu/~rus
 |9|354|Demodulation|[Demodulation](/aima-core/src/main/java/aima/core/logic/fol/inference/Demodulation.java)|
 |9|354|Paramodulation|[Paramodulation](/aima-core/src/main/java/aima/core/logic/fol/inference/Paramodulation.java)|
 |9|345|Subsumption|[SubsumptionElimination](/aima-core/src/main/java/aima/core/logic/fol/SubsumptionElimination.java)|
+|9.8|341|Append|---|
+|10.1|369|AIR-CARGO-TRANSPORT-PROBLEM|---|
+|10.2|370|SPARE-TIRE-PROBLEM|---|
+|10.3|371|BLOCKS-WORLD |---|
+|10.7|380|HAVE-CAKE-AND-EAT-CAKE-TOO-PROBLEM|---|
```
samagra14, line 88: These are not AI algorithms as such. These are the problems to which they can be applied. In my humble opinion, these should not be added to the list of algorithms. Maybe a separate list of implemented environments and/or questions should work.

ritwik12: @samagra14 I think adding them to the list will help everyone to know which algorithms are still left. So that they may not waste their time in searching for them. just like I did. I don't want others to do the same.

ritwik12: I think aimacode wants a single list of readme. If they agree with another list, then your opinion is right.

norvig:
There could be a separate "Index of Data Structures" at the bottom of the README file, like in aima-python https://github.com/aimacode/aima-python#index-of-data-structures

ritwik12:
@norvig Sure

PR #400, opened 2018-04-03
Fixes typos and makes Java 8 related changes to the explicit types for ArrayLists (FOL AIMA4e)

samagra14:
This PR makes minor changes to fix typos in method names as well as comments. This PR also removes unwanted explicit types of ArrayLists.


samagra14:
@norvig Please have a look here.

core/src/main/java/aima/core/logic/basic/firstorder/Unifier.java
```diff
@@ -104,7 +104,7 @@ public Unifier() {
 			// else if x = y then return theta
```
norvig, line 104: I really don't think you need these comments. They are just echoing the code.

PR #401, opened 2018-04-04
FOL-FC-ASK implementation

samagra14:
This PR adds the implementation of the FOL-FC-ASK algorithm. This PR adds FOLFcAsk.java with the following dependencies.
* An interface to represent the FOL knowledge Base
* An interface to represent Clauses

I have added an interface to represent the FOL knowledge base so that alternate implementations of the KnowledgeBase can be added to the aima.extra module by simply implementing the interface. Also, this will make it easier to implement custom KnowledgeBases. Besides this, the implemented algorithm makes very little assumptions regarding the organisation of the KnowledgeBase.

However, I will be adding a concrete implementation of the KnowledgeBase for testing the algorithm. The KnowledgeBase interface, however, is not exhaustive. It will become richer as implementations of further algorithms are added.

### DIFF
--- a/core/src/main/java/aima/core/logic/api/firstorderlogic/FOLKnowledgeBase.java
+++ b/core/src/main/java/aima/core/logic/api/firstorderlogic/FOLKnowledgeBase.java
@@ -1,0 +1,38 @@
+package aima.core.logic.api.firstorderlogic;
+
+import aima.core.logic.basic.firstorder.kb.data.Literal;
+import aima.core.logic.basic.firstorder.parsing.ast.Sentence;
+import aima.core.logic.basic.firstorder.parsing.ast.Term;
+import aima.core.logic.basic.firstorder.parsing.ast.Variable;
+
+import java.util.List;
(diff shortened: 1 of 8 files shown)


samagra14:
@norvig 
Please have a look here.

core/src/main/java/aima/core/logic/api/firstorderlogic/FOLKnowledgeBase.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @return A list of all the definite implication Clauses useful for inference
+     *      algorithms.
+     */
+    List<Clause> getAllDefiniteClauseImplications();
+
+
+    /**
+     * Used to check if a sentence is a <b>renaming</b> of a known fact. One sentence is a
```
norvig, line 39: I'm not sure it is worth worrying about renaming (especially since the default implementation ignores the issue). COuldn't we do without this complication?

PR #404, opened 2018-05-02
Implementation of Planning Algorithms

samagra14:
This is a Work in Progress PR. Please don't merge this PR. In this PR I plan to implement chapter 10 algorithms. [This](https://samagra14.github.io/blog/2018/05/02/AI-Planning.html) is a simultaneous blog on the implementation. I will be regularly adding commits to this PR.

- [x] Add Parser
- [x] Add applicability check
- [x] Add Problem Factory
- [x] Add GraphPlan Algorithm
- [ ] Add comments and test cases


aima-core/src/main/java/aima/core/logic/planning/PlanningProblemFactory.java
```diff
@@ -0,0 +1,56 @@
...
+                "^At(P2,JFK)^Cargo(C1)^Cargo(C2)^Plane(P1)^Plane(P2)^Airport(JFK)^Airport(SFO)");
+        State goalState = new State("At(C1,JFK)^At(C2,SFO)");
+        Variable c = new Variable("c");
+        Variable p = new Variable("p");
+        Variable a = new Variable("a");
+        Variable from = new Variable("from");
+        Variable to = new Variable("to");
+        ArrayList variables = new ArrayList<>(Arrays.asList(c, p, a));
+        ArrayList flyVars = new ArrayList<>(Arrays.asList(p, from, to));
+        ActionSchema loadAction = new ActionSchema("Load", variables,
```
norvig, line 21: This is good as is, but maybe it would be easier if ActionSchema took a string, not a ArrayList, as the second argument? Here the string would be "c, p, a", and we wouldn't need the 7 declarations of variables and ArrayLists.

samagra14: Yeah, strings are a lot better. I will improve this in the next PR.

PR #410, opened 2018-05-25
Implements updated FOL-BC-ASK

samagra14:
This pull request implements updated FOL-BC-Ask. I have not included generator like constructs as Java doesn't support yield keyword directly. If generators were to be implemented then the algorithm would have to be split into various methods of the `Iterable ` interface in java. This implementation bears the closest resemblance with the pseudocode and hence I have implemented it in this way.

**Typical generator in java**
````java
// Iterator implemented as anonymous class.  This uses generics but doesn't need to.
for (int i: new Iterable<Integer>() {
    @Override
    public Iterator<Integer> iterator() {
        return new Iterator<Integer>() {
            int counter = 1;

            @Override
            public boolean hasNext() {
                return counter <= 100;
            }

            @Override
            public Integer next() {
                return counter++;
            }

            @Override
            public void remove() {
                throw new UnsupportedOperationException();
            }
        };
    }
}) {
    System.out.println(i);
}
````


samagra14:
@norvig Please have a look.

PR #411, opened 2018-05-28
Hierarchical search

samagra14:
Implementation of hierarchical search.
 Travelling to San Francisco airport problem is used as a test case. This algorithm cannot be used with the spareTire problem as the branching factor of the spareTire problem becomes too high when implemented in the form of a single base HLA (Act).


samagra14:
@norvig Please have a look.

PR #413, opened 2018-06-02
Adds contributing guidelines

samagra14:
@norvig Please have a look at these contribution guidelines. I have tried to be as specific as possible. Do let me know if there is anything else that can be added.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,82 @@
+Contributing to AIMA-JAVA
+==========================
+:+1::tada: First off, thanks for taking the time to contribute! :tada::+1:  
+If you'd like to report a bug or join in the development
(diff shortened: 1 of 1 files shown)


ritwik12:
@samagra14 Looks good!!

norvig:
The commit messages, `:racehorse`, etc. seem strange to me. Is there established practice to using these names? I don't think I could memorize them, and most of them I wouldn't recognize when I read them, unless I went back and looked at the key.

samagra14:
These instructions were mutual to many popular FOSS projects and therefore I included them. Those commit messages are optional and can be used to increase the readability of the commits. These can be removed if necessary.

norvig:
If these are common, then we should leave them in. They are just new to me.

PR #416, opened 2018-06-04
Version space learning

samagra14:
Adds version space learning. Instead of evaluating and storing the entire hypothesis space, I have utilised the partial order property of hypotheses in the hypothesis space. Hence, I have implemented version spaces by storing only two sets of hypotheses: a most general boundary set (the G-set) and a most specific boundary set (the S-set).


samagra14:
@norvig Please have a look.

norvig:
Looks good!

PR #420, opened 2018-06-22
Ontological Engineering

samagra14:
This notebook demonstrates how to represent the aspects of the real world using first-order logic. In this notebook, I am trying to show the stepwise implementation of such a system. There is a scope of adding other logics such as Modal and Descriptive Logics. However, they will require changes to the logic module. These can be targeted as future issues.


samagra14:
@norvig 
Currently, we don't have a module for modal logic. However, there is no pseudocode or algorithm which requires modal logic to be implemented, but I feel that it is a beautiful concept. I was wondering if it was worth the time to add a small module for working with modal logic. It would be a great issue to take up post-GSoC.
Should I open it as an issue? Or should I leave it for the time being?

samagra14:
@norvig 
Please have a look.

norvig:
Nicely done.

PR #425, opened 2018-06-26
Aima gui improvements

samagra14:
1. Corrections to fol bc ask to support GUI apps.
2. Reverts to the older test suite for FOLBC.
3. FOLBC is made to implement the inference procedure.


samagra14:
@RuedigerLunde @norvig 
This PR solves the problem of GUI apps. I am sorry for the delay as I wanted to make sure all the work done till now was compatible with all the modules of the project. Also, I have reverted to the previous version of the testing suite for this algorithm so that the architecture remains the same. Please have a look.

PR #431, opened 2018-07-08
AIMA4e Search

samagra14:
This PR adds the generic search algorithm to the AIMA4e branch. For testing, the generic search is extended as bfs and dfs. It is then applied to the Romania problem.


samagra14:
@norvig Please have a look.
 Should I remove Tree search and graph search from AIMA4e or should they coexist alongside generic search?

core/src/main/java/aima/core/search/basic/uninformedsearch/BreadthFirstSearch.java
```diff
@@ -10,31 +10,79 @@
...
+        //  if problem's initial state is a goal then return empty path to initial state
         if (problem.isGoalState(problem.initialState())){
             return nodeFactory.newRootNode(problem.initialState());
         }
+        // frontier ← a FIFO queue initially containing one path, for the problem's initial state
         Queue<Node> frontier = new LinkedList<>();
         ((LinkedList<Node>) frontier).add(nodeFactory.newRootNode(problem.initialState()));
+        // reached ← a set of states; initially empty
         HashSet<S> reached = new HashSet<>();
+        // solution ← failure
```
norvig: I don't like this style, where the comments duplicate the code. I think it is good to have the pseudocode in a block comment, but then the Java implementation should stand by itself. Any Java programmer knows what `reached = new HashSet<>();` means; we don't need to duplicate that in the comments.

samagra14: @norvig 
Sure, sir, I will have these duplicate comments removed. I added them just because of the older versions. I will have them removed.

samagra14:
@norvig 
This PR concludes all the modified algorithms.
1. I have removed a lot of unwanted interfaces(for eg SearchController etc).
2. Implemented informed search from scratch as opposed to extending uniform cost search.
3. AndOrGraphSearch and the algorithms after that have not been modified in the draft and hence I have not modified their implementation. All other algorithms have been modified.
5 Removed unwanted comments.
6. In a few algorithms, I have kept the older testSuite as it was quite exhaustive.
Please review this PR.

samagra14:
@norvig 
Is there anything else that needs to be corrected and/or added in this PR?

norvig:
Looks good!

Issue #459, opened 2019-05-17
GSOC 2019: ND4J and DL4J exploration


samagra14:
Explore the use cases of ND4j and write a short notebook to aid the user in understanding it's usage in the project.

samagra14:
@jsuyash1514 Are you planning to add something extra or are you planning to carry this notebook along with the 18th chapter?

jsuyash1514:
What I'm planning is, we can have separate notebooks for the 18th and 19th chapter. This notebook is just to demonstrate the usage of nd4j and dl4j libraries. As an example, suppose while implementing the 19th chapter, we perform an operation on a matrix, let's say a matrix multiplication using nd4j, so we can add a brief illustration of this step (i.e. the use of nd4j in performing a matrix multiplication), in this notebook. 
This notebook will, therefore, aid the user in understanding the essentials of these libraries and hence their usage in the project.
@samagra14 What's your opinion on this? May I proceed the same way?

jsuyash1514:
@samagra14 I'll keep on adding to this notebook simultaneously as we go on implementing the 18th chapter.

PR #460, opened 2019-05-19
[WIP] ND4J and DL4J notebook

jsuyash1514:
This PR initiates a notebook which serves as a user guide for ND4J and DL4J libraries.
Solves #459 
Till now I've added:-
1.) A brief explanation of "What these libraries are used for?"
2.) Different ways to create an N-Dimensional Array.
Future plans:- 
As the implementations of 18th and 19th chapters of the textbook would require some mathematical operations on the N-Dimensional arrays, I'll keep on adding the ways to perform these operations using these libraries in the notebook, so that the user can refer to this notebook later, to understand the usage of these libraries in the project.


jsuyash1514:
@samagra14 Please review.

PR #473, opened 2019-12-15
Small modification to DecisionTreeLearner

UditArora2000:
Replaced iterator with lambda function


RuedigerLunde:
Thank you for your proposal. I have now used a slightly different solution.

PR #545, opened 2026-06-09
:art: Add seed support to MazeVacuumEnvironment for reproducible runs

### LINKED ISSUE #544: MazeVacuumEnvironment does not support reproducible runs
When comparing different agent strategies in the Maze environments, each run produces a different layout and agent starting position because the randomness is seeded from `System.currentTimeMillis()`.
This makes it impossible to do a fair, repeatable comparison between agents without manually fixing the environment state in code.

Proposed change:

- Add a seeded constructor to MazeVacuumEnvironment so callers can pass an explicit long seed to reproduce the same maze layout and dirt distribution.
- Expose a seed input in VacuumAgentApp so users can optionally fix the seed from the GUI, including the agent's initial placement.

**Expected behaviour:** 
When a seed is set, re-running the experiment with any agent produces the exact same maze, dirt pattern, and starting position, enabling a fair side-by-side comparison.

---
*Issue text drafted with the assistance of generative AI (Claude by Anthropic).*

Jakob-42:
# Closes #544

### MazeVacuumEnvironment

* Stores a `java.util.Random` instance (seeded) as a field instead of calling the stateless `Util.randomInt()`.
* Adds a new constructor:

  ```java
  MazeVacuumEnvironment(int xDim, int yDim, double dirtProbability, double obstacleProbability, long seed)
  ```

  which accepts an explicit seed.
* Preserves the existing two-argument and four-argument constructors.
* Updates the four-argument constructor to delegate to the new seeded constructor using `System.currentTimeMillis()` as the seed, ensuring existing call sites remain unaffected.

### VacuumAgentApp

* Adds a **Use Seed** checkbox and a seed `TextField` (default value: `404`) to the toolbar.
* When **Use Seed** is unchecked (default), behavior remains unchanged and the seed is derived from `System.currentTimeMillis()`.
* When **Use Seed** is checked, both the maze layout and the agent's initial placement are generated using the supplied seed, guaranteeing an identical starting state across runs and across agents.

## How to test

1. Select **Small Maze Environment** or **Maze Environment**.
2. Check **Use Seed** and leave the value as `404`.
3. Run an agent (`RandomWalkVacuumAgent`) and note the maze layout and starting square.
4. Click **Init** and run again.
5. Verify that the maze layout, dirt pattern, and starting square are identical.
6. Uncheck **Use Seed**.
7. Verify that each **Init** generates a new random maze.

## Checklist

* [x] Existing constructors and call sites are unaffected (backward-compatible)
* [x] No new dependencies introduced
* [x] GUI seed control is disabled by default (opt-in only)

---

*PR description drafted with the assistance of generative AI (Claude by Anthropic).*

### DIFF
--- a/aima-core/src/main/java/aima/core/environment/vacuum/MazeVacuumEnvironment.java
+++ b/aima-core/src/main/java/aima/core/environment/vacuum/MazeVacuumEnvironment.java
@@ -53,9 +53,12 @@ public class MazeVacuumEnvironment extends VacuumEnvironment {
 
 	@Override
 	public void addAgent(Agent<? super VacuumPercept, ? extends Action> agent) {
-		super.addAgent(agent);
-		if (envState.getLocationState(getAgentLocation(agent)) == null)
-			envState.setLocationState(getAgentLocation(agent), LocationState.Clean);
+		List<String> freeLocs = new ArrayList<>();
+		for (String loc : getLocations())
+			if (!containsObstacle(loc))
+				freeLocs.add(loc);
+		List<String> candidates = freeLocs.isEmpty() ? getLocations() : freeLocs;
+		addAgent(agent, candidates.get(random.nextInt(candidates.size())));
 	}
 
 	public void setObstacle(String location, boolean b) {
--- a/aima-core/src/main/java/aima/core/environment/vacuum/MazeVacuumEnvironment.java
+++ b/aima-core/src/main/java/aima/core/environment/vacuum/MazeVacuumEnvironment.java
@@ -3,7 +3,6 @@ package aima.core.environment.vacuum;
 import aima.core.agent.Action;
 import aima.core.agent.Agent;
 import aima.core.agent.impl.DynamicAction;
-import aima.core.util.Util;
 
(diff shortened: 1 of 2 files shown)


RuedigerLunde:
Thank you very much for this suggestion; it really makes sense to me. One issue is that `addAgent` can place the agent on an obstacle. The current implementation of `MazeVacuumEnvironment` uses a workaround here and simply removes the obstacle in such cases. If we were to adapt this addAgent implementation (it could use a seed-based random placement), that would also have the advantage of requiring only a single `Random` instance. What do you think?

Jakob-42:
Great point — I've moved the seeded placement into `MazeVacuumEnvironment.addAgent(Agent)` itself, using the already-stored random field. The new implementation collects all non-obstacle locations first and picks from those, so the obstacle workaround is gone entirely. VacuumAgentApp is back to a plain env.addAgent(agent) with no instanceof check needed. Thanks for the suggestion!

RuedigerLunde:
Thank you so much. Now it's perfect!

